PR #11, opened 2018-05-29
Proteomes

4ndr01d3:


src/components/Entry/DomainArchitectures/index.js
```diff
@@ -167,37 +175,39 @@ class DomainArchitectures extends PureComponent {
...
+                      }}
+                    >
+                      {' '}
+                      {d}{' '}
+                    </Link>{' '}
+                    -{' '}
                   </span>
                 ))}
-                <IDAProtVista matches={idaObj.domains} length={idaObj.length} />
+                <IDAProtVista matches={idaObj.domains} length={1000} />
```
aurel-l: Why was that changed to a fixed value, is that temporary?

4ndr01d3: Yes, it is now constant because the domains don't have length, so everything is now calculated on an hypothetical length of 1000, however I should've  use the previously defined constant length. (line 61)

aurel-l: Ah, OK, makes sense. Do you want to do the change to use a common const value?

4ndr01d3: Done!

PR #63, opened 2019-02-06
Structure ngl

CodeGit:


gustavo-salazar:
I have added the requested changes. I guessed @CodeGit is going to be too busy! 😄

PR #101, opened 2019-07-03
Taking dev to the Future

gustavo-salazar:


src/components/Description/index.js
```diff
@@ -16,165 +16,191 @@ import fonts from 'EBI-Icon-fonts/fonts.css';
 
 const f = foundationPartial(ebiStyles, styles, theme, fonts);
 
-export const ParagraphWithCites = ({
```
CodeGit: It might be worth renaming ParagraphWithCites to ParagraphsWithCitations for search-ability in future.

src/components/Description/index.js
```diff
@@ -16,165 +16,191 @@ import fonts from 'EBI-Icon-fonts/fonts.css';
...
+                target="_blank"
+                className={f('ext')}
+                key={part}
+              >
+                {tagValue}
+              </Link>
+            );
+          }
         }
         // TODO: change the way descriptions work from the backend side.
```
CodeGit: Good idea, it makes sense for much of this to happen in the API

PR #159, opened 2019-10-03
Pfam residue conservation

CodeGit:
Improved the conservation popup


gustavo-salazar:
I got an error with the protein I was testing.
The API returns some `"None"` values for the `aa` key. See https://www.ebi.ac.uk/interpro/api/protein/unreviewed/A0A015IN71/?conservation=
And that brakes the logic choosing the `hit.color`

src/components/Related/DomainsOnProtein/index.js
```diff
@@ -30,44 +30,70 @@ const ProtVista = loadable({
...
-      color = '#2451a4';
-    } else {
-      color = '#00429d';
-    }
-    /* eslint-enable no-magic-numbers  */
+    const hit = colourMap.find(element => {
+      return residue.score > element.min && residue.score <= element.max;
+    });
 
+    const color = hit.color;
```
gustavo-salazar: This is the line that fails when the API returns None. because hit is empty.

PR #192, opened 2019-12-06
Reordering ida

gustavo-salazar:
Improvements on the Search by IDA:

* The domains can be dragged to reorder them.
* A domain that matches an accession, is painted with the color of it
* Changing the display of pfam or InterPro accessions is now in the redux state
* Minor style fixes.

### DIFF
--- a/src/components/SearchByIDA/style.css
+++ b/src/components/SearchByIDA/style.css
@@ -94,7 +94,7 @@ ul.ordered {
   //border: black solid 2px;
   border-radius: var(--entry-padding);
 
-  @nest & button {
+  @nest & button.close {
     position: absolute;
     background: transparent;
     font-size: 0.8em;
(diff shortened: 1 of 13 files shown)


src/components/SearchByIDA/style.css
```diff
@@ -104,9 +105,33 @@ ul.ordered {
     //border: black solid 1px;
     width: 1em;
   }
-  @nest & button:hover {
+  @nest & button.drag {
```
CodeGit: I suggest putting draggable sections on both the beginning and the end of the domain graphic. I would also consider centre justifying the text in the domain graphic or increasing the amount of padding between the draggable section and the domain text.

gustavo-salazar:
@CodeGit about your comments:
* Not so sure about adding a second draggable section I think it would look funny
* between centering or adding padding I vote for padding.
* I'll make the suggested changes in the text

PR #193, opened 2019-12-12
Notifications section added in settings 

Swaathik:


gustavo-salazar:
I like it! 

But I wonder if we can abstract its usage to avoid some repeated code... Maybe creating a new component Tip, that has as props: `{title, body, checkLabel, settingLabel}` and the addToast gets call from there. That way we have a single point for code that is currently repeated, for example the `updateToastSettings` methods.

Maybe @CodeGit or lorna can suggest edits on the text... not only the content of the Toast but also the name of the setting category.

I don't think the tip to click on the InterProScan ID is actually necessary. Seems intuitive to me.

PR #196, opened 2019-12-17
Submit annotation

Swaathik:


gustavo-salazar:
I started by checking the commit on the API and I have a couple of suggestions in that side:
* When an error is [returned](https://github.com/ProteinsWebTeam/interpro7-api/blob/541a19e1fd2c5f616abcdff62370082c55768dd8/webfront/views/mail.py#L25) because of too many attempts, it should return a HTTP code error `429` Too many requests.
* maybe name the method [`store_credentials`](https://github.com/ProteinsWebTeam/interpro7-api/blob/541a19e1fd2c5f616abcdff62370082c55768dd8/webfront/views/mail.py#L42) something like `store_credentials_and_mail`
* Add an HTTP code status for the error in the [last line](https://github.com/ProteinsWebTeam/interpro7-api/blob/541a19e1fd2c5f616abcdff62370082c55768dd8/webfront/views/mail.py#L71)

gustavo-salazar:
And for the client side, I think what is missing is the error handling.
And maybe show a toast, acknowledging the mail sent.

PR #200, opened 2020-01-20
Zoom overlay

Swaathik:


gustavo-salazar:
Just found bug, panning by dragging either the tree or the Protvista component is not working.
Not sure if is at the level of the nightingale component or the way we are dealing with it in interPro7.

gustavo-salazar:
I guess is at nightingale level because the error happens here as well: https://ebi-webcomponents.github.io/nightingale/#/manager

PR #204, opened 2020-01-28
Changes related to the use of /wwwapi

gustavo-salazar:


src/store/utils/get-initial-state/index.js
```diff
@@ -42,6 +42,10 @@ export default history => {
       };
     }
   }
+  // TODO: remove this line when the logs show that users are now using /wwwapi/
+  if (settings?.api?.hostname === 'www.ebi.ac.uk') {
```
CodeGit: When is this code executed? The very first time a browser visits the site, or everytime there's a new session? some users may leave months or years between visists to interpro.

It might be worth adding api.root="/api/" to the condition to trigger the code below so that we only change the settings when the setting is wrong.

gustavo-salazar: the code gets executed when the page is refreshed or loaded for the first time.
I have put a reminder in slack so we can remove this line in a few months.

At this moment the settings is only an object and hasn't been loaded so I dont think it is too expensive to execute it everytime.

PR #225, opened 2020-04-23
InterProScan Result Page

Swaathik:
The changes include 

- The renaming of the page title itself.

- Allowing users to edit the title of the InterProScan search. In this Pull request, it is implemented only in the main page where the user can browse/view the result.

- To let the users know that the search js going on, a spinning circle is shown while the job is running.


src/components/IPScan/Summary/index.js
```diff
@@ -243,8 +244,9 @@ const changeTitle = (
 ) => {
   if (inputRef.current.value !== '') {
     results.xref[0].name = inputRef.current.value;
+    const input = `>${inputRef.current.value} ${results.sequence}`;
```
gustavo-salazar, line 247: I wonder if the lack of a new line `'\n'` in between the name and the sequence might cause any side effects... I check around and seems ok but worry if we are missing something

src/components/IPScan/Summary/index.js
```diff
@@ -379,7 +365,24 @@ const SummaryIPScanJob = ({
                         Update
                       </button>
```
gustavo-salazar, line 366: Do you think a cancel button would be useful... in case you don't want to change the title any more?

src/components/IPScan/Summary/index.js
```diff
@@ -342,20 +342,6 @@ const SummaryIPScanJob = ({
...
-                        'button-space',
-                      )}
-                      data-icon="&#xf303;"
-                      aria-label={'Edit Title'}
-                      onClick={() => setShowTitleField(!showTitleField)}
-                    />
-                  </Tooltip>
                   {showTitleField ? (
                     <div className={f('title-form')}>
                       <input
```
gustavo-salazar, line 347: Can you prepopulate this field with the current name. it might be the case the user just want edit a current name

src/components/IPScan/Summary/index.js
Swaathik: Good point. As we are completely hiding the old title, it is worth adding a button to go back to what it was

src/components/IPScan/Summary/index.js
Swaathik: The input field in idb when the job is created for the first time is shown the same way. I think its fine. I checked if it is elsewhere. Seems okay to me

src/components/IPScan/Summary/index.js
Swaathik: yeah, Sure. I will make the change

gustavo-salazar:
Couple of comments:
* please change the focus to the `<input>` when it changes to writeable.
* set a `min-width` for the `<input>`, maybe something like `12em`, so it shows some space if the chosen title is too short
* I notice you are reusing the same `<button>` to make the `<input>` editable, that's cool but maybe change the icon from ✏️ to 💾 or something to show it's use to save.

PR #247, opened 2020-07-21
Reorganising protvista options

Swaathik:
All the options are now moved to the top. 

There is one thing we shall discuss on using the dropdown - whether to close the dropdown automatically when user selects options like Snapshot and Print?

The rest of the options I feel it would be unnecessary to close them without letting the users to play around.


gustavo-salazar:
hey @Swaathik I added a couple of details...
1. embedded the option in the menu as `<label>` so in that way the user can click on the text to trigger the event.
2. Change a bit the logic of the `[accession]: [name]` labels, using a more react-like approach.
3. The Protvista file has grown a lot, I was wondering if it would be worthy to separate the logic of the Options into its own component.

PR #262, opened 2020-09-03
Curation tab added to Pfam entries

Swaathik:
Curation has to be added in `config.yml ` under entry subpages.

### DIFF
--- a/src/subPages/Curation/index.js
+++ b/src/subPages/Curation/index.js
@@ -1,0 +1,143 @@
+// @flow
+import React from 'react';
+import T from 'prop-types';
+
+import Loading from 'components/SimpleCommonComponents/Loading';
+import { createSelector } from 'reselect';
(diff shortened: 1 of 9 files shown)


src/subPages/Curation/index.js
```diff
@@ -0,0 +1,153 @@
...
+              Sequence: {details.hmm.cutoffs.gathering.sequence}
+              <br />
+              Domain: {details.hmm.cutoffs.gathering.domain}
+            </td>
+          </tr>
+          <tr>
+            <td>Download</td>
+            <td>
+              <Link
+                href={`${config.root.API.href}/entry/pfam/${payload?.metadata?.accession}?annotation=hmm`}
```
gustavo-salazar, line 102: I changed the link to get the hmm from our API to reduce the use of pfam website

PR #270, opened 2020-09-30
Show favourites

Swaathik:


src/components/Title/index.js
```diff
@@ -314,22 +336,57 @@ class Title extends PureComponent /*:: <Props> */ {
...
+                    role="button"
+                    onClick={() => this.manageFavourites(metadata)}
+                    onKeyDown={() => this.manageFavourites(metadata)}
+                    tabIndex={0}
+                  >
+                    <span
+                      className={f(
+                        'icon',
+                        'icon-common',
+                        this.props.entries.includes(metadata.accession)
```
gustavo-salazar, line 367: Changed this way as I think it's a bit more react-like...
this way the dom element doesn't change only its css class.

Although I'm not actually sure if the virtual DOM takes care of it

src/components/home/FavouriteEntries/index.js
```diff
@@ -0,0 +1,101 @@
...
+                {this.state.fav.map((f) => (
+                  <EntryCard entry={f} key={f.metadata.accession} />
+                ))}
+              </AnimatedEntry>
+            </div>
+          </div>
+        </div>
+      );
+    }
+    return (
```
gustavo-salazar, line 78: Just added a bit more text to inform about how to favourite entries

src/storage/idb/index.js
```diff
@@ -18,6 +19,9 @@ const init = () => {
         case 0:
           db.createObjectStore(IPScanJobsMeta, { autoIncrement: true });
           db.createObjectStore(IPScanJobsData, { autoIncrement: true });
+        // eslint-disable-next-line no-fallthrough
+        case 1:
```
gustavo-salazar, line 23: So this way, when someone that already has a version of InterPro running, gets only the update of the version 2.

gustavo-salazar (approved):

src/components/Title/index.js
Swaathik: Oh right. I was just getting rid of the errors and warnings at the time which I forgot to refactor it later.

PR #279, opened 2020-10-13
Ipscan from file

gustavo-salazar:
With this PR you can upload an InterProScan result file to visualize the results in the website.

This is a good example as it has all expected results/analyse: `/ebi/production/interpro/nuka/interproscan-5/distributions/release/interproscan-5.46-81.0/output/test_all_plus.run2.dp.json`

The whole Import area is set to capture dropping files so users can just drag their icon there, as an alternative for the upload button.

### DIFF
--- a/src/actions/creators/index.js
+++ b/src/actions/creators/index.js
@@ -174,6 +174,10 @@ export const importJob = (job /*:  Object */) => ({
   type: types.IMPORT_JOB,
   job,
 });
+export const importJobFromData = (job /*:  Object */) => ({
+  type: types.IMPORT_JOB_FROM_DATA,
+  job,
+});
 
 export const unloadDataJob = (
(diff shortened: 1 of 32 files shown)


CodeGit (approved):

CodeGit:
I added an extra paragraph on the status page to direct users towards the import button

gustavo-salazar:
Thanks @CodeGit 
I have also received some feedback from Gift, that I'll be working on. So I'll add some more changes before we merge this PR

gustavo-salazar:
Hey @CodeGit and @Swaathik  I have added some stuff requested by gift.
Please check everything still works in this PR. 
The new changes:
* The name of the job is taken from the sequence id
* Clear All jobs... as some files can have hundreds of sequences
* Checks if the job was run with the current version of Interpro scan, and warns the user if is not the case.

PR #280, opened 2020-10-14
Hmmersearch

CodeGit:


gustavo-salazar (approved):

gustavo-salazar:
@CodeGit I added a commit to split the hmmer button as a separate component.
Have a look and merge if you are happy with it

CodeGit:
I've added a couple of attributes to the button to allow the message and size to be customised. What do you think?

gustavo-salazar:
Yeah cool. Merging it now

PR #281, opened 2020-10-16
Browser notifications

Swaathik:
Both the sequence search and download support browser notifications and there is an option to enable them in the settings page.
I have removed the click event for download as it caused page refresh and the redux state for download is lost. If there is any way to retain it, please direct me. I will add it.

### DIFF
--- a/src/components/DownloadForm/Controls/index.js
+++ b/src/components/DownloadForm/Controls/index.js
@@ -59,7 +59,7 @@ export class Controls extends PureComponent /*:: <Props> */ {
 
   _handleGenerateClick = blockEvent(() => {
     // Request browser notification
-    askNotificationPermission(() => {});
+    askNotificationPermission();
 
     this.props.downloadURL(
       this.props.url,
--- a/src/components/DownloadForm/Controls/index.js
+++ b/src/components/DownloadForm/Controls/index.js
@@ -9,6 +9,7 @@ import NumberComponent from 'components/NumberComponent';
 
 import { downloadURL, downloadDelete } from 'actions/creators';
 
+import { askNotificationPermission } from 'utils/browser-notifications';
 import blockEvent from 'utils/block-event';
 import { toPlural } from 'utils/pages';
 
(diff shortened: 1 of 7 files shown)


CodeGit:
I've tried viewing the popup in firefox and chromium and get the following errors in the console after running an interproscan search:

Screenshot_20201016_160118
Screenshot_20201016_160842

Any ideas what might be going wrong here?

src/components/DownloadForm/Controls/index.js
```diff
@@ -56,14 +57,17 @@ export class Controls extends PureComponent /*:: <Props> */ {
     noData: T.bool.isRequired,
   };
 
-  _handleGenerateClick = blockEvent(() =>
+  _handleGenerateClick = blockEvent(() => {
+    // Request browser notification
+    askNotificationPermission(() => {});
```
gustavo-salazar, line 62: Minor comment, but I think It's kind of weird to have to pass an empty function. An alternative is to use `noop` from lodash as the default so if you don't provide `setState` it calls `noop`.

I did something like it here: https://github.com/ProteinsWebTeam/interpro7-client/blob/b3b6cdead118b886e163764bdfc262d3d54253df/src/components/SimpleCommonComponents/Modal/index.js

gustavo-salazar:
I't looking good!
I would add similar logic to ask for permission somewhere in `components/File` because you could trigger a download job from there:
Screenshot 2020-10-19 at 16 48 07

src/pages/Settings/index.js
```diff
@@ -119,100 +120,136 @@ const NotificationSettings = (
...
+                settings.
+              </p>
+              <button className={f('button')} disabled>
+                Notifications {Notification.permission}
+              </button>
+            </>
+          ) : (
+            <>
+              <p>
+                Allow browser to notify when the InterProScan search has
```
gustavo-salazar, line 155: Not sure of the wording here. Maybe something like:

Allow us to send you browser notifications when one of your Jobs or Downloads finishes.

@CodeGit what do you think?

src/components/DownloadForm/Controls/index.js
Swaathik: Thanks for the comment Gustavo. I will change it.

src/pages/Settings/index.js
```diff
@@ -119,100 +120,136 @@ const NotificationSettings = (
...
+      />
+      <div className={f('row')}>
+        <div className={f('medium-12', 'column')}>
+          {status === 'answered' ? (
+            <>
+              <p>
+                To change notification preference, please do in the browser
+                settings.
+              </p>
+              <button className={f('button')} disabled>
```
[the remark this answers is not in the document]
Swaathik: I made quite a few changes on this part. Now I have made it an infobox to show the message. Please have a look! If we could settle on the message for both cases, it would be great

gustavo-salazar (approved):

Issue #282, opened 2020-10-16
Display issue when leaving full screen mode of the protein viewer in structure pages


typhainepl:
When going in to view the protein viewer in full screen mode on structure pages and existing again, the viewer shows up on the right hand side of the 3D structure instead of being displayed below it (see image attached).
![image](https://user-images.githubusercontent.com/25660657/96255896-98ab7900-0faf-11eb-8ef5-3583e9afc5fa.png)

PR #284, opened 2020-10-20
fixing bug with fullscreen in structure page

gustavo-salazar:
Re-factory of the fullscreen logic in the structure page after finding bug reported in #282


Swaathik (approved):

Swaathik:
I think the minimize button was deleted accidentally. I have included it in the control bar when the viewer is stuck at the bottom. Please check if is okay

gustavo-salazar:
Thanks @Swaathik ! I removed that completely by mistake!

PR #287, opened 2020-10-28
Storybook

Swaathik:
The items mentioned in [here](https://www.ebi.ac.uk/panda/jira/secure/RapidBoard.jspa?rapidView=48&projectKey=IBU&view=planning&selectedIssue=IBU-9352&epics=visible&issueLimit=100&selectedEpic=IBU-8930) have been added. 

For the Link example, the internal link is yet to be done. Since the redux state is a dummy one, the custom location prop of `Link` does not work.


stories/InfoBlock.stories.js
```diff
@@ -0,0 +1,28 @@
...
+
+const f = foundationPartial(ebiGlobalStyles, fonts);
+
+export default {
+  title: 'Basic UI/InfoBlock',
+};
+
+export const Basic = () => (
+  <div
+    className={f('callout', 'info')}
```
gustavo-salazar, line 15: It would be cool to have the different styles of callouts: https://get.foundation/sites/docs/callout.html#

Maybe use a select knob (https://github.com/storybookjs/storybook/tree/master/addons/knobs) as we are already using knobs (e.g MemberDBSelector)

stories/ToggleSwitch.stories.js
```diff
@@ -0,0 +1,26 @@
...
+import ebiGlobalStyles from 'ebi-framework/css/ebi-global.css';
+import fonts from 'EBI-Icon-fonts/fonts.css';
+
+const f = foundationPartial(ebiGlobalStyles, fonts);
+
+export default {
+  title: 'Basic UI/Switch',
+};
+
+export const Basic = () => (
```
gustavo-salazar, line 13: I think this is a good candidate to create a React component for.

stories/Link.stories.js
```diff
@@ -0,0 +1,53 @@
...
+    <Link href="https://www.ebi.ac.uk/interpro"> Interpro</Link> /
+    <Link href="https://www.ebi.ac.uk/" className={f('ext')} target={'_blank'}>
+      {' '}
+      EMBL-EBI
+    </Link>
+  </>
+);
+
+export const InternalLink = () => {
+  // TODO yet to be fixed
```
gustavo-salazar, line 46: The way I did it in the MemberDBSelector story was to import the component without the connect so instead of importing `Link` import `_Link` (you gonna have to expose that in the Link component.)
That way you can pass the `setLocation` as a mock for `goToCustomLocation`

stories/InfoBlock.stories.js
Swaathik: It's a cool idea. All the styles can be exposed including the close button.

stories/GOTerm.stories.js
```diff
@@ -0,0 +1,36 @@
...
+    category: {
+      code: 'C',
+      name: 'cellular_component',
+    },
+    identifier: 'GO:0042025',
+    name: 'host cell nucleus',
+  },
+];
+export const Basic = () => (
+  <GoTerms terms={data} type={'entry'} db={'interpro'} />
```
gustavo-salazar, line 35: maybe display this with different values for type and db... or maybe use knobs?

stories/FullScreen.stories.js
```diff
@@ -0,0 +1,29 @@
...
+import FullScreenButton from '../src/components/SimpleCommonComponents/FullScreenButton';
+
+import { foundationPartial } from '../src/styles/foundation';
+import fonts from 'EBI-Icon-fonts/fonts.css';
+const f = foundationPartial(fonts);
+export default {
+  title: 'Basic UI/FullScreen',
+};
+
+export const Basic = () => {
```
gustavo-salazar, line 11: The FullScreenButton has a two attributes to _hook_ functionality before entering to fullscreen mode and after exiting... I think it would be useful to have a second story that uses them for something basic... maybe changing the background color.

stories/ClanViewer.stories.js
```diff
@@ -0,0 +1,115 @@
...
+    },
+    {
+      source: 'PF12661',
+      target: 'PF12662',
+      score: 0.0051,
+    },
+  ],
+};
+
+export const Basic = () => {
```
gustavo-salazar, line 107: Out of the scope of this PR, but one thing we haven't done is for the clan viewer to support the [ctrl] to zoom style that we have in protvista, I'll create an issue for it

stories/Skylign.stories.js
```diff
@@ -0,0 +1,80 @@
...
+
+export default {
+  title: 'InterPro UI/Skylign',
+};
+const data = {
+  alphabet: 'aa',
+  max_height_theory: 6.453114921518104,
+  max_height_obs: 5.2996166606681,
+  min_height_obs: 0,
+  height_arr: [
```
gustavo-salazar, line 14: Maybe generate/replicate `height_arr` and `probs_array` so it has enough length that scrolling over the component is necessary

stories/GOTerm.stories.js
Swaathik: I have been looking for examples for this, any suggestion?

gustavo-salazar: Mhh... you got me here... it seems that it looks the same with different values of `type`:
* https://www.ebi.ac.uk/interpro/protein/UniProt/A0A000/
* https://www.ebi.ac.uk/interpro/entry/InterPro/IPR004839/
Where the only difference is the Tooltip.

And I check in the database and the only db with go terms is `interpro`, so not sure how useful is the `db` attribute.

Swaathik: yes, I did check for all the other DBs. If it is going to be only interpro [this check ](https://github.com/ProteinsWebTeam/interpro7-client/blob/master/src/components/GoTerms/index.js#L65) is not necessary right

gustavo-salazar: Not at the moment. But some member DB have their own GO annotations, is only that we aren't importing them now. 
So I would leave that code there, just in case later we decide to include their go annotations.

stories/Link.stories.js
gustavo-salazar: Your code remind me of the mocked redux store that is been using in connected table, so I created a ConnectedLink using that strategy.
Also added a `<pre>` area showing the current state of the location

gustavo-salazar (approved):

PR #294, opened 2020-11-06
Isoforms improvements

gustavo-salazar:
A few improvements when displaying protein isoforms:

* The selected isoform is now included in the URL (e.g. `protein/UniProt/P31749/?isoform=P31749-2`) and corresponding part of the redux state (e.g. `{customLocation: {search: {isoform: 'P31749-2'}, ...}, ...}`)
* The order and groups on the protvista graphic for an isoform is similar to the one for a protein. Only missing residues and extra features because that data is not available for isoforms
* When an isoform is selected, it could be displayed fullscreen together with the protein for comparison purposes.

### DIFF
--- a/src/components/Protein/Isoforms/Selector/index.js
+++ b/src/components/Protein/Isoforms/Selector/index.js
@@ -6,6 +6,7 @@ import { format } from 'url';
 import { createSelector } from 'reselect';
 import loadData from 'higherOrder/loadData';
 import descriptionToPath from 'utils/processDescription/descriptionToPath';
+import { goToCustomLocation } from 'actions/creators';
 
 import Loading from 'components/SimpleCommonComponents/Loading';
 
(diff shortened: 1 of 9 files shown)


Swaathik (approved):

Swaathik:
There is just one thing(Not related to the code changes above), when colour by option is selected in one viewer, the selection is updated in the other viewer's options but not reflected in the viewer. 

The rest of the options are fine. they behave only when the respective options are changed. Since it is coming from the `settings`, it behaves different

gustavo-salazar:
Do you mean in the Isoform protein viewer? or in all protein viewers?

gustavo-salazar:
Ok I manage to replicate the issue... I'll check if is a quick fix... maybe just connecting the component... and if is more complicated than that I'll open a ticket

Swaathik:
In places where there is more than one protein viewer like in https://www.ebi.ac.uk/interpro/structure/PDB/1a0h/. Yes it makes sense to open a new ticket as it is unrelated to this PR.

PR #297, opened 2020-11-12
Taxo subpage changes

Swaathik:


gustavo-salazar:
The request for `api//taxonomy/uniprot/protein/entry/InterPro/IPR000001/?extra_fields=counters&page_size=20` still happens  even if the setting selection is only _Key Species_.
I think it is requested at https://github.com/ProteinsWebTeam/interpro7-client/blob/f58d9efe8dd4e95c29fa3b24ce960e3e8164e774/src/components/Related/index.js#L449

I think you manage to get it as a _non blocker_  and `<KeySpeciesTable>` will load even if the request in Related is not completed, but the query is still executed. 

Do you think that request can be avoidable without major refactoring of this code?

gustavo-salazar:
And about the reorganization of the keySpecies table, maybe remove the links on the Tax ID, name and protein counts columns and add another action to go to the species page.
And for the icons use the same ones that we have for counters in the Cards, for instance the ones in 'Latest Entries'

Swaathik:
Hey @gustavo-salazar, please check the last commit and let me know if it is a good strategy to take it forward.

gustavo-salazar (approved):

gustavo-salazar:
I just added a small validation in the tooltip because the same table column is used in the proteome subpages.
Otherwise, I'm happy with your solution.

PR #299, opened 2020-11-20
GitHub actions

gustavo-salazar:
Travis has changed their policies.

1. [LINK](https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom) `.org` will disappear and all OS project will get move to `.com`
2. [LINK](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) change in their paying plans. Giving only a once off set of credits to use in the free plan, once that runs out you need to move into their pay plan.

### Description of changes
I have setup the same test CI tasks task that we had in travis but using github actions. You can see the runs coming from the branch here: https://github.com/ProteinsWebTeam/interpro7-client/actions


Swaathik (approved):

Swaathik:
I went through the documentation to understand more clearly about the steps and workflow terms. It looks great

PR #300, opened 2020-11-25
Update favourites locally

Swaathik:
The client notifies when there are changes in type, name, member DBs, description, literature and Go terms when a new version is detected. It is implemented to update the IndexedDB content even if no difference is detected in the above properties. As the favourite entries in the home page takes the counters from the browser's storage, it has to be updated anyway.

I haven't included the modal window in this PR for the above reason. Please correct me if I am getting it wrong.


gustavo-salazar:
I checked the code and it seems alright for the notification. 
But I think this misses a couple of things:

1. What if the user doesn't have the notifications enabled, We can add a Toast for it in that case
2. But also you manage to identify the changes in your compare function, but this, I think would be useful to show to the user; on a minimal level saying something like : `This items have changed: name and literature` and on a more advanced level showing a diff view or a before and after, for each of the fields of the changed entries.
And that's why I was talking of an overlay message, but maybe we can have a page for this, and the notification redirects to that page. 
The page can show the diff if the version in indexedDB is different to the current version. other wise shows a message like `All your favourite entries are up to date.`

gustavo-salazar:
Also not sure if is a problem... but if a user have 10 favourites, then is going to receive 10 notifications. Maybe we can group all of them in a single notification.

src/utils/compare-favourites/index.js
```diff
@@ -0,0 +1,120 @@
...
+  notify = false,
+  addToast = noop,
+  setChangedFav = noop,
+}) => {
+  const favTable = await getTableAccess(FavEntries);
+  const storedContent = await favTable.getAll();
+  const entries = Object.entries(storedContent);
+  const changedEntries = [];
+  entries.forEach(([key, value], index) => {
+    fetch(`https://www.ebi.ac.uk/interpro/api/entry/interpro/${key}`).then(
```
gustavo-salazar, line 70: Get the API URL from the settings

src/pages/FavouriteUpdates/index.js
```diff
@@ -0,0 +1,80 @@
...
+                    splitView={true}
+                    leftTitle={'Saved'}
+                    rightTitle={'Latest'}
+                    hideLineNumbers={true}
+                  />
+                </div>
+              ))}
+            </>
+          ) : (
+            <h2>All your favourites are up to date</h2>
```
gustavo-salazar, line 72: This appears while the data is been collected from indexedDB, which for a second gives the impression there are not differences... 
Find a way to show a message for "Looking for the changes in your favourite entries."

src/pages/FavouriteUpdates/index.js
```diff
@@ -0,0 +1,80 @@
...
+                            setChangedFav,
+                          )
+                        }
+                      >
+                        {' '}
+                        Update
+                      </button>
+                    </div>
+                  </div>
+                  <ReactDiffViewer
```
gustavo-salazar, line 60: I like how it looks. And I tested the colors using [chrome emulation of  vision deficiencies](https://developers.google.com/web/updates/2020/03/devtools#vision-deficiencies) and I think the color contrast is good!

I wouldn't show the JSON in the difference but rather each of the values you are checking for the difference. But I think you were planning to do that anyway.

src/utils/compare-favourites/index.js
```diff
@@ -0,0 +1,120 @@
...
+      (response) => {
+        if (response.ok) {
+          response.json().then((json) => {
+            const differences = compare(value.metadata, json.metadata);
+            if (Object.keys(differences).length > 0) {
```
gustavo-salazar, line 75: Is this alright?
I think there is an issue here... differences is an object that even when they are equal looks like:
```
{
  newData: {},
  old: {},
}
```
So it will always have keys... or am I missing something?

src/components/home/FavouriteEntries/index.js
```diff
@@ -70,6 +71,16 @@ export class FavouriteEntries extends PureComponent /*:: <Props> */ {
                   <EntryCard entry={f} key={f.metadata.accession} />
                 ))}
               </AnimatedEntry>
+              <Link
```
gustavo-salazar, line 74: Not sure if we should show this.... or maybe we can add a validation to only show if there are differences.

gustavo-salazar:
Despite the appearances from all my comments, I really like how this is looking, good job! 👏🏽 👏🏽 👏🏽

src/utils/compare-favourites/index.js
Swaathik: :see_no_evil: That is the part I changed a lot. I knew there must be some blunder. Thanks for pointing it out 👍

.flowconfig
```diff
@@ -21,6 +21,8 @@
 .*/node_modules/react-popper*
 .*/node_modules/react-hotkeys*
 .*/node_modules/react-select*
+.*/node_modules/create-emotion*
```
Swaathik, line 24: @gustavo-salazar I tried installing separately but the list goes endlessly after every `npx flow-typed install .....`. I have added in the ignore list. I also noticed that the package `@emotion` is included in the list - https://github.com/ProteinsWebTeam/interpro7-client/blob/master/.flowconfig#L19.

It was the same causing the errors after installing create-emotion.

gustavo-salazar:
I did a couple of minor changes in 441a9a5 : 
* Ask for permission to send notifications if a new favourite is added
* Fix a bug in `compare-favourites:41` where the base was `null` 

A bit of a bigger change was done in `compare-favourites:165` where I use promises to make sure all the entries have been compared before calling the callback. The previous approach there was inconsistent.

Have a look and if you think is OK, delete the testing code and and merge.

gustavo-salazar (approved):

PR #302, opened 2020-12-07
Minor ipscan changes

gustavo-salazar:
A few changes related in the interproscan results area:

1. Save in Browser moved to the actions row in the result's summary.
2. New state for a job: `saved in browser` to differentiate wwwprod jobs that have been saved from loaded files.
3. The mismatched version warning has been updated to include a message when there is a new version of InterProData and wwwprod hasn't updated interproscan.


CodeGit (approved):
Looks fine, although I think we should remove all reference to "job" from any messages to users and refer to them instead as search or results (or search results) for interproscan searches.

PR #305, opened 2021-01-07
Taxonomy name exact match

CodeGit:


src/pages/Taxonomy/index.js
```diff
@@ -747,39 +751,63 @@ const getURLFromState = createSelector(
...
+          hostname,
+          port,
+          pathname: root + descriptionToPath(desc),
+        });
+      } catch {
+        return;
+      }
+    } else if (search && search.match(/^[\w ]+$/) && description.taxonomy) {
+      const desc = {
+        main: {
```
gustavo-salazar, line 774: Not having `...description` here, makes the query to lose any other endpoints that are been filtered by. So the `scientific_name` searches over all the taxonomy names.
This might not be a bad thing because it allows to show inner tax nodes.

But it causes undesired side effects, like the counters will include more entries/proteins.

For example, check the human entry here:
http://localhost:8000/interpro/taxonomy/uniprot/entry/smart/?search=sapiens#table
v.s.
http://localhost:8000/interpro/taxonomy/uniprot/entry/smart/?search=homo%20sapiens#table

The problem is that it seems that the API is not supporting the `scientific_name` search for multiple endpoints, as for example this query fails:
http://localhost:8007/api//taxonomy/uniprot/entry/interpro?scientific_name=Homo%20sapiens

CodeGit:
I've made the changes necessary to get the correct numbers in the exact match row.

gustavo-salazar (approved):

PR #306, opened 2021-01-11
Ipscan minor changes

Swaathik:
* It includes the show/disable button changes for the export button in InterProScan result page. For locally stored results, if they are deleted in our server (past 7 days), only JSON can be downloaded and other options are greyed out.
* Search with InterProScan option is added to protein page

### DIFF
--- a/src/components/IPScan/Summary/index.js
+++ b/src/components/IPScan/Summary/index.js
@@ -254,6 +254,7 @@ const SummaryIPScanJob = ({
   remoteID,
   localTitle,
   status,
+  times,
   data,
   localPayload,
   api,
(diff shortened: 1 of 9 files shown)


src/components/IPScan/Summary/index.js
```diff
@@ -329,6 +330,17 @@ const SummaryIPScanJob = ({
 
   const goTerms = getGoTerms(payload.matches);
 
+  let dataURL = 'https://www.ebi.ac.uk/Tools/services/rest/iprscan5/result/';
+  const now = Date.now();
+  const expired =
```
gustavo-salazar, line 335: Instead of using `now` as the failover of `created` maybe use something similar to [line 314](https://github.com/ProteinsWebTeam/interpro7-client/blob/4228c4fb5d15f1deb7987dc5106be18cc14f82b7/src/components/IPScan/Summary/index.js#L314) where the creation date is infer from the ipscan ID.

gustavo-salazar (approved):

src/components/IPScan/Summary/index.js
gustavo-salazar: or not sure if we can actually do that calculation when the job is been imported, just to make sure the `created` value is always there

Swaathik: When it is imported 'created' value is undefined. But the whole things needn't be calculated for imported jobs as we are disabling the export anyway.

gustavo-salazar: but it's exactly at the point of deciding it should be disabled or not that it's important to have it available. 
In the current setup the 7 days are counted from the moment it's imported, but the job could have been imported 6 days after it was created, so it actually has only 1 day more in the server.

gustavo-salazar (approved):

PR #307, opened 2021-01-20
Search terms stored in local storage and displayed in home page

Swaathik:
When the user types and the search is triggered, thats when the search term is stored in local. So the same DEBOUNCE RATE is used. 

My question is when the user makes a selection from the search results, it has to be stored too right? Like they type a keyword, and goes for a particular entry, that entry would be his recent search isn't it?

### DIFF
--- a/src/components/IPScan/Summary/index.js
+++ b/src/components/IPScan/Summary/index.js
@@ -248,6 +248,7 @@ const getEntryURL = ({ protocol, hostname, port, root }, accession) => {
   });
 };
 
+// eslint-disable-next-line complexity
 const SummaryIPScanJob = ({
   accession,
   localID,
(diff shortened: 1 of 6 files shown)


gustavo-salazar:
Cool, I think most of the functionality is there. ONe thing I think is missing is to be able to remove a particular entry. I would add a `X` at the side of the tag to be able to remove that single entry.

About styling I would use the `tag` class similar to the `InterPro Entry` in this pic (maybe a different color, but in that style)
Screenshot 2021-01-20 at 12 46 27

And the n the list of searched terms can be similar to the way https://stackoverflow.com/ show a list of tags:
Screenshot 2021-01-20 at 13 00 16

Swaathik:
Im actually working on the tag style now. It would save more space too. Cool. I will add the functionality to remove a single item

gustavo-salazar (approved):

PR #308, opened 2021-01-29
Filtering and Sorting added

Swaathik:


gustavo-salazar:
So there a few things to fix before we can merge this. 

The biggest issue is that If the filters cause an empty result the whole component fails, and the page is now empty. So the user cannot change what caused the empty result. The page should be able to display the table with no results, like when we search in the browse area - [example](https://www.ebi.ac.uk/interpro/entry/InterPro/?search=void#table)

Here are other comments:
* confidence is a numeric value only, so maybe use `<input type="number"`, and it would be nice to show a `>` to show that the number will filter with anything larger or equal to the input.
* Given that the protein will only work with complete matches, we can use the uniprot regex, to validate the field, and only when is a valid accession we actually send the query... maybe changing the color of the input to let the user know that is valid or not. Also maybe including a Tooltip explaining that only uniprot accessions are valid search terms.
* Sometimes the sorting doesn't work.  if I'm in http://localhost:8000/interpro/entry/InterPro/IPR000001/genome3d/?sort_by=&evidences=pDomTHREADER&accession=Q9V6K3#table and then sort by confidence the order doesn't change. but if I refresh that page, then it does.

src/components/Genome3D/List/index.js
```diff
@@ -84,6 +86,15 @@ export const List = (
   customLocation: {search?: Object}} */,
 ) => {
   if (data.loading) return <Loading />;
+  if (!data.payload)
```
gustavo-salazar, line 89: I was thinking on something like this but inside the table, so all the search controllers are still there and the user can change the selection that took it into an empty set

src/components/Genome3D/List/index.js
```diff
@@ -157,6 +157,10 @@ export const List = (
               </Link>
             )}
             isSearchable={true}
+            customiseSearch={{
+              validation: /^([OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z][0-9]([A-Z][A-Z0-9]{2}[0-9]){1,2})$/i,
```
gustavo-salazar, line 161: It seems that genome3d is still not supporting case insensitive, maybe we should `toUpperCase` in the URL when you are searching by this field.

src/components/Genome3D/List/index.js
```diff
@@ -157,6 +157,10 @@ export const List = (
...
+              message: 'Accepts only UniProt accessions',
```
gustavo-salazar, line 162: It took me a while to find where this message was appearing, maybe using a tooltip that displays an icon at the side of the textbox is more obvious

gustavo-salazar (approved):

PR #309, opened 2021-02-03
Browse redesign

gustavo-salazar:
Redesign of the Browse area. 
===

Currently deployed in http://wp-np3-aa.ebi.ac.uk/future/

The changes included are:

* All filters moved to the left side 
* `MemberDBSelector` as the top filter in the left side. It is scrollable when not enough space.
* Endpoint tabs removed. Instead the Browse Item in the main menu is now a dropdown list so the user has to show what data is browsing by.
* Dropdown menus for other items of the main menu (e.g. About)
* Adding Breadcrumbs to show the current place in the website.
* Browse `By Entries` is now split into `By InterPro` and `By Member DB`
* EBI visual framework updated to 1.4. Taking it from github for now, but I was told that a release on npm should be available this week.
* Improvements on resolution filter on Browse By Structure
* FTP links updated to https in the Download page
* Replacing react-autocomplete for react-select
* Removing memberDBselector for browse by proteome


src/components/SearchByText/TextSearchBox.js
```diff
@@ -186,4 +198,6 @@ const mapStateToProps = createSelector(
   (main, value, pageSize) => ({ main, value, pageSize }),
 );
 
-export default connect(mapStateToProps, { goToCustomLocation })(TextSearchBox);
+export default connect(mapStateToProps, { goToCustomLocation }, null, {
+  forwardRef: true,
```
Swaathik, line 202: ForwardRef here refers the passed Ref to this component right. The same thing but more in a react way?

src/components/SearchByText/TextSearchBox.js
```diff
@@ -50,24 +53,25 @@ class TextSearchBox extends PureComponent /*:: <Props, State> */ {
     inputRef: T.func,
```
Swaathik, line 53: InputRef can be removed as it is not used. This is what forwardRef is doing right?

Swaathik (approved):

src/components/SearchByText/TextSearchBox.js
gustavo-salazar: Sort of... not necessarily more react, but sort of force to do it that way because of the connect.
The thing is that I need the reference of the [react component](https://github.com/ProteinsWebTeam/interpro7-client/blob/45966b1f68789b3e0c1b370a867f1d951987fa9d/src/components/SearchByText/index.js#L78) to be able to call a method from it (`focus` in this case). But if I  do it the normal way with  normal `ref`, React complains that Functional components are not allowed to have references, the funny thing is that `TextSearchBox` is not funcional, it is defined using the `class` syntax. but because it is wrapped with `connect` and `connect` is now implemented using hooks, means that the wrapper is a functional component, hence the error.
So the people from `react-redux` added the ForwardRef option so they add a ForwardRef in the wrapper and then you can put  a `ref` in the component.

src/components/SearchByText/TextSearchBox.js
gustavo-salazar: Thanks indeed, this is leftover from before I found the solution via ForwardRef

PR #332, opened 2021-03-31
Panther conservation

Swaathik:
- Conservation URL is changed to fetch scores for PANTHER entries.
- The fetched scores are used to calculated the average based on a window size of 25.
- The color scale is replaced with a line graph track. The track is shown for only one PANTHER entry (the longest if many entries are hit).

- [ ] @CodeGit Can you please check if the tooltip we display [here](https://github.com/ProteinsWebTeam/interpro7-client/blob/panther_conservation/src/components/ProtVista/index.js#L471) is right and also the logic [here](https://github.com/ProteinsWebTeam/interpro7-client/blob/panther_conservation/src/components/Related/DomainsOnProtein/index.js#L424) is still valid as the HMMsearch is used now?


CodeGit:
We should probably hide the conservation button for unreviewed proteins as the API can't generate Hmmer data for those proteins

PR #339, opened 2021-04-28
Heatmap for contacts distance matrix in structural model page

Swaathik:


CodeGit (changes requested):
I'm having trouble viewing the model for PF06980

http://localhost:8000/interpro/entry/pfam/PF06980/model/

Is it possible that the memory requirements for the heatmap are too hight. I saw the chrome renderer requesting over 3GB of ram. Could you check on your laptop?

CodeGit (approved):

PR #360, opened 2021-10-18
new hover listener

CodeGit:
This should create a short label in the viewer when hovering over a protein residue. It works but appears to gradually degrade performance. I'm not sure whether this is due to the way I'm handling the events or if it's specific to me laptop (which is rather old and struggles running VScode and devtools)


gustavo-salazar:
@CodeGit I just put the code where the event is subscribed inside a `useEffect` that only executes when the `viewer` changes, that way the event is only listened once. I think that prevents the memory leak you were experiencing. Please try it and let me know.

PR #370, opened 2022-02-09
Taxonomy sunburst + dependencies update

gustavo-salazar:
Basically has 2 parts:
1. The new sunburst visualisation for taxonomies. Which also includes a bit of reorganization to make it the default view and disable the general setting for displaying table or key species.
2. Dependencies updates

**Note 1:**  The sunburst info is a comes from a new modifier in the API so it requires https://github.com/ProteinsWebTeam/interpro7-api/pull/71 if you want to test this locally.

**Note 1:** Sorry didn't know this was this big of a PR 🙈
But fear not! `136` of the files are just updates to flow types, so if flow runs without errors, that should be enough of a test for those files; so you actually just need to check `23`! much easier isn't?


src/components/Related/Taxonomy/Sunburst/style.css
```diff
@@ -0,0 +1,33 @@
...
+    display: flex;
+    align-items: center;
+    word-break: keep-all;
+
+    & input {
+      margin-left: 0.7em;
+      margin-right: 0.7em;
+    }
+  }
+}
```
matthiasblum, line 33: The extension is `.css`, but it looks like SCSS. Any reason for not name the file `style.scss`?

matthiasblum (approved):
Well done!

src/components/Related/Taxonomy/Sunburst/style.css
gustavo-salazar: It's because we are not using sass there, it is actually postcss, and it gets process in webpack together with normal css.

matthiasblum: I see. Thank you for the clarification. :)

PR #373, opened 2022-02-24
Improving performance in Lighthouse

gustavo-salazar:
Multiple changes to the bundle with the purpose of improving the performance of the site. This will be mostly noticeable for slow internet connections on their first load (as the service worker will help for subsequent connections).

Here are the changes:
* Lighthouse:
  * Removing custom script to run lighthouse and using lhci instead
  * results of lighthouse in CI will be included in github pull requests
  
* Crawling script to generate the static content of the homepage and inject it into the `index.html` (_a la SSR_). **Not in use**, as I wasn't able to make react hydrate that content, even when using hydrate it discards it and recreates it again. 🤷🏽  I'm leaving the script around, as it could still be a good strategy if the hydration works.

* Script to use [Critical](https://github.com/addyosmani/critical), which gets the critical css and injects it in the `dist/index.html`. **Not in use**, as it didn't show significant improvement, and it was too cumbersome to run. Leaving the script here, as I think it might be useful if the mentioned pre-generation of content works.

* Changes in the `index.template.html` : 
  * Support for the a webpack bundle that hydrates instead of rendering.
  * Tags for fonts use now`woff2` as those are the fonts that actually been used
  * `preconnect` to some of the servers we are getting resources from
  * The `<body>` now has temp divs to emulate the headers
  * Some styling for those block to have size and color of what will be rendered eventually.
  * `<script async`
* Changes in `webpack.config.js`:
  * Adding `index.hydrate.js` as a second entry point. **Not in use**, as this is only useful if using the pre-generated content.
  * Adding optimization plugin `splitChunks` to try to split any big chunks that were not explicitly requested via dynamic imports.
  * Bit of a tweak on the babel target to try to avoid replication in the 2 bundles.
  * Removing compression libraries as now zlib is native in node, so it can be used here instead.
  * Including source maps in production, as several performance test suites can make use of them for their analyses, e.g. lighthouse does, and I used `source-map-explorer` to quickly check the content of the bundle files.

* Some changes in the code to bust the performance:
  * Remove the general import of the big css files: foundation, ebi-global, new-interpro, etc. As they are been dynamically imported around the app, and therefore they were added twice.
  * The header are now imported dynamically to create an extra chunk and avoid the inclusion of the big styling files in the main chunk.
  * Dynamically importing a few libraries so they also get their own chunk, e.g. `yaml-js` and `pako`
  * Some CSS specificity changes needed to correct undesired side effect of not importing the libraries globally.
  * Using `/* webpackPreload: true */` for the import of `pages` and `Home`, so they get loaded asap and that way improve the LCP time
  * Grouping all the dynamic imports of the memberDB logos so they create a single chunk instead of tiny ones.


aurel-l:
General question regarding the generating content + hydration path that has not been followed:
I understand hydration didn't work because react tries to match the existing DOM tree to what it would generate and whenever it finds something different it bails out and just replace the whole thing.
Have you tried to just not even try hydrating, and just render _in place of_ the static content? Like at the moment there's a static loading indicator within the root of the application to be displayed before react could kick in, there could be a bit more than this (for example, basic header + footer + loader in the main area)?
I know this is still extra work for the browser that will be discarded by react's rendering, but at least:
 - the user sees something a bit more useful (hopefully the links in the header and footer could still work as they're plain links)
 - react wouldn't _try_ to do the matching, so it would be less wasteful than hydrating then bailing

src/pages/index.js
```diff
@@ -22,7 +23,11 @@ const f = foundationPartial(ebiGlobalStyles, ipro, style);
 
 // Main pages
 const Home = loadable({
-  loader: () => import(/* webpackChunkName: "home-page" */ './Home'),
+  loader: () =>
+    import(
+      /* webpackPreload: true */
```
aurel-l, line 28: I'm wondering what's the proportion of users that enter the website not through the home page but through entry pages directly (following external links, or a Google search). Because for those then this preload might be a waste of resources that might never actually be used.

gustavo-salazar: About the `webpackPreload` for the `Home` component, it is about 12% of all visits that go into the homepage, 5% to the search page, and the rest of pages are around 1% or much less. So I do think the unnecessary load in other pages is justifiable. Moreover, thanks to other changes in this PR the main bundled wa reduced to at least half of what it was, so other pages are going to get a booster in performance nonetheless.

gustavo-salazar: Thanks for sharing the `lazy.ts` code I'll read it properly and see if we can do something similar!

src/components/SearchByText/TextSearchBox.js
```diff
@@ -4,12 +4,12 @@ import T from 'prop-types';
 import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
 import { debounce } from 'lodash-es';
+import loadable from 'higherOrder/loadable';
```
aurel-l, line 7: More linting than anything, I guess this is not really useful here if the corresponding element is commented out
```suggestion
// import loadable from 'higherOrder/loadable';
```

src/index.template.html
```diff
@@ -127,54 +170,72 @@
...
+        // Adding js files for module entry point
+        for (const file of htmlWebpackPlugin.files.moduleScripts || []) {
+          if (file.indexOf(htmlWebpackPlugin.entry)!==-1){
+            %> <script async type="module" src="<%= file %>"></script> <%   
+          } 
+          if (file.indexOf('.hydrate.')===-1 && file.indexOf('.index.')===-1){
+            %> <script async type="module" src="<%= file %>"></script> <% 
+          } 
+        }
+        // Adding js files for legacy entry point
```
aurel-l, line 229: Unrelated to the PR itself, but I'd be interested in the proportion of users using the legacy modules instead of the more "modern" bundle if you have this number somewhere

aurel-l (approved):
Looks good in general. Interesting gotchas to take into account so that's interesting to see. Looking forward to see the improvement in the deployed website 👍🏽

src/subPages/index.js
```diff
@@ -4,15 +4,13 @@ import { format } from 'url';
 import loadable from 'higherOrder/loadable';
 import loadData from 'higherOrder/loadData';
 import descriptionToPath from 'utils/processDescription/descriptionToPath';
+// Too small to be split.
+import Proteome from './Proteome';
```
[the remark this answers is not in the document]
gustavo-salazar: Yeah, I think the `minSize` feature works only when the chunk is dynamically created by webpack, but when using a dynamic import it chunk it no matter its size, I eventually discovered that there is a _magic comment_ to indicate that you don't want to split in an specific `import()` but can remember the exact syntax from the top of my head

src/index.template.html
gustavo-salazar: I don't know an easy way to get that number exactly. But I did a quick calculation of the browser usage for the month of February in analytics, and cross referencing it with `caniuse` info on which version support esmodules, gave an approximate number of 2.5% of people using InterPro that is still using legacy browsers.
I'm guessing this fluctuates a lot, for example, in the data I was looking, there was a random spike on firefox 44, that justified almost half the people using the legacy bundle.

PR #380, opened 2022-04-13
Side menu instead of tabs

gustavo-salazar:
Changing the tabs of an entity page into a collapsible left side menu. 
In mobile the menu is displayed on top of the content and it's not collapsible

Side changes:
* Simplifies the animation of the underline of the active item in the menu
* Options to collapse all filters or clear all filters are now only shown when there is more than one filter.
* The width of the taxonomy tree component only updates on intervals of `100px` this to avoid a weird bug where in certain conditions it kept growing indefinitely. Min width is now 600px
* Makes sure that links to `/protein/uniprot` are tagged as active when the current page is `/protein/reviewed` or `/protein/unreviewed`


matthiasblum (approved):
Looks good.

I wish the EBI framework would not force a `max-width` on the `row` class, so we could use the full width of the screen (+ padding), but that's unrelated to this PR.

Feel free to merge (and sorry it took me so long to review :confounded:)

gustavo-salazar:
> I wish the EBI framework would not force a `max-width` on the `row` class, so we could use the full width of the screen (+ padding), but that's unrelated to this PR.

Oh believe me, for that and many other reasons I would really like to upgrade to the new version of the framework and not been forced into foundation, but that I think is a task too big for the benefits that would bring! 🤷🏽

PR #381, opened 2022-04-14
Grouping PIRSR into PIRSF entries

gustavo-salazar:
Adds custom logic to display PIRSR residues:

* If a PIRSR residue(e.g. `PIRSR00025-1`) matches the accession of a PIRSF family (e.g. `PIRSF00025`) with a match in the current protein, it will be integrated there as residues of it, similar to what is done to CDD residues.
* If it doesn't it will be kept as Other residues
* The other residues block can be collapsed by default.

Additionally the collapse button is a bit more obvious in feature viewer.

Deployed in the future environment for testing. e.g. http://wp-np3-aa.ebi.ac.uk/future/protein/reviewed/D3ZTE0/ or http://wp-np3-aa.ebi.ac.uk/future/protein/reviewed/P43403/


tgrego (approved):
Looks great, Residues display is spot on.
The Other Residues might benefit in the future from a merge similar to what is done in the Residues, but I would not bother much with it given that in my opinion the important site matches are the ones coming from the family (PIRSF) match.

All great!

matthiasblum (approved):
Looks good.

I am still puzzled by site annotations without a PIRSF family match, but in the meantime the page looks much better.

Thank you! :+1:

PR #385, opened 2022-04-27
Short name in feature viewer

gustavo-salazar:
It adds the option `Short Name` to the labels in the feature viewer.

The options are now kept in the redux state instead of the component state, to make this option permanent and be reflected in future uses of the feature viewer.

The option can now also be change in the settings page.

Bonus commit: 💅 Styling the settings page

*note*: I'll open a different PR to deal with the short names in the clan viewer

### DIFF
--- a/src/pages/Settings/index.js
+++ b/src/pages/Settings/index.js
@@ -1,5 +1,5 @@
 // @flow
-import React, { PureComponent, useState } from 'react';
+import React, { PureComponent, useState, useEffect } from 'react';
 import T from 'prop-types';
 import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
@@ -18,7 +18,12 @@ import { installPrompt } from 'main';
 
 import loadable from 'higherOrder/loadable';
 
-import { changeSettings, resetSettings, addToast } from 'actions/creators';
+import {
+  changeSettings,
+  changeSettingsRaw,
+  resetSettings,
+  addToast,
+} from 'actions/creators';
 import { askNotificationPermission } from 'utils/browser-notifications';
 
 import { EntryColorMode } from 'utils/entry-color';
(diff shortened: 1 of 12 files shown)


src/pages/Settings/index.js
```diff
@@ -603,7 +699,34 @@ class _AddToHomeScreen extends PureComponent /*:: <Props,State> */ {
...
+    label,
+  } /*: { firstVisibleTitle: string|null, hash:string, label: string } */,
+) => (
+  <li>
+    <Link
+      to={(current) => ({
+        ...current,
+        hash,
+      })}
+      href="#settings-notification"
```
matthiasblum, line 715: Shouldn't it be `href={'#' + hash}`?

matthiasblum (approved):
Look good. It's really nice to have a general setting for this.

One minor issue I found in the setting page is if you click on a link in the new menu on the left, it move to the correct section, but if you click again on the same link, it moves to a different section.

src/pages/Settings/index.js
gustavo-salazar: Oh  I think that actually the error is that `href` that is still there. It's redundant because I'm defining the `to`. 
I'll play a bit with this.

gustavo-salazar:
> Look good. It's really nice to have a general setting for this.
> 
> One minor issue I found in the setting page is if you click on a link in the new menu on the left, it move to the correct section, but if you click again on the same link, it moves to a different section.

I think the underlying issue is related to the stickiness of the main header, there is an offset that needs to be taken into account and I believe that is causing the weird behaviour. I'll take a look into it

gustavo-salazar:
@matthiasblum I changed the way that the settings menu works, instead of keeping up to date with the global stat via `<Link>` I made it a local state thing, so I can keep track of the scroll better.
This should correct the issue that you reported. let me know if you can have a look to it

Issue #386, opened 2022-05-09
Incorrect reprensation of InterPro match


matthiasblum:
When submitting a sequence to InterPro (https://www.ebi.ac.uk/interpro/search/sequence/), InterProScan only reports signature matches (e.g. Pfam, PANTHER, etc.) but not InterPro matches. The InterPro tracks are added by the website.

However, there is an bug for multi-domain matches: the InterPro track covers the entire region instead of the domains reported by InterProScan.

**Example 1: A0A075FKY9**

InterPro search result

![image](https://user-images.githubusercontent.com/2097501/167432159-8a21ef52-39f0-4042-a6be-8cdee72e608e.png)

InterPro (https://www.ebi.ac.uk/interpro/protein/UniProt/A0A075FKY9/)

![image](https://user-images.githubusercontent.com/2097501/167433057-45ef1adf-818a-4ce3-b105-39ac337e89c5.png)

**Example 2: A0A009ECR1**

InterPro search result

![image](https://user-images.githubusercontent.com/2097501/167434564-a725f475-a06e-4642-8b9c-eb767b5ed5ab.png)

InterPro (https://www.ebi.ac.uk/interpro/protein/UniProt/A0A009ECR1/)

![image](https://user-images.githubusercontent.com/2097501/167435087-abe7f16b-e501-4219-b5ec-be4774fc3610.png)

**Example 3: A0A009E983**

InterPro search result

![image](https://user-images.githubusercontent.com/2097501/167435369-d0208e77-4073-45c1-bd4c-e1a45317585c.png)

InterPro (https://www.ebi.ac.uk/interpro/protein/UniProt/A0A009E983/)

![image](https://user-images.githubusercontent.com/2097501/167435243-1c340e30-3aae-47d4-87a3-765bb97e2dc8.png)

PR #390, opened 2022-06-01
Iprscan re construction of IPR locations

gustavo-salazar:
As reported in #386, the representation of InterProScan results was different to the one shown in a protein page that has the exact same sequence.

The difference was due to the way the locations for InterPro entries were calculated in the client, which differs from the way are calculated in production. Basically, production creates 2 locations out of 2 non-overlapping signature matches, while the client was creating a single location that covered both.

This PR updates the client to behave in the same way as the production algorithm.

--

Additionally, I extracted some of the logic into a different file to separate what is component logic and what is data mangling.

### DIFF
--- a/src/components/IPScan/Summary/StatusTooltip/index.js
+++ b/src/components/IPScan/Summary/StatusTooltip/index.js
@@ -1,5 +1,7 @@
-import React, { useEffect, useState } from 'react';
+// @flow
+import React from 'react';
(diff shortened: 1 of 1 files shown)


matthiasblum (approved):
Looks good to me. 

I tried to submit a few sequences, and I always got consistent results. Thank you!

Issue #387, opened 2022-05-10
Download scripts write FASTA sequences more than once


matthiasblum:
Original issue: ebi-pf-team/interproscan#266

When downloading in the FASTA format using the generated code snippets (example: [eukaryotic proteins matched by IPR008160](https://www.ebi.ac.uk/interpro/result/download/#/protein/UniProt/entry/InterPro/IPR008160/taxonomy/uniprot/2759/|fasta)), sequences with more than one hit are written several times because match locations are part of the header, e.g.

```
>A0A016RS24|IPR008160|159...215|Col_cuticle_N domain-containing protein
MDSDHRLKAYRFVAYSAVTFSVVAVLSVCVTLPMVYNYVHHVKMSVRSEVRHCKKTAGDIWGEVTHIKNIPTANRTARQS
GYGGDGEVQGNSNYGNNGGNGGSCSGCCMPGPPGGAGAPGRPGKPGRPGAAGMPGNPGKSGGSPCDPKTPPPCKPCPAGE
PGPAGPPGPPGDAGPDGAPGNPGYGSIPGEPGPKGPPGPDGQPGYPGSPGQPGQPAYSTPPSPGEPGPQGPPGPQGPPGP
DGQPGYGGGQGQPGPKGPPGPDGQPGPDGNPGQPGYPGQPGGPGEPGICPKYCAIDGGVFFEDGTRR
>A0A016RS24|IPR008160|225...283|Col_cuticle_N domain-containing protein
MDSDHRLKAYRFVAYSAVTFSVVAVLSVCVTLPMVYNYVHHVKMSVRSEVRHCKKTAGDIWGEVTHIKNIPTANRTARQS
GYGGDGEVQGNSNYGNNGGNGGSCSGCCMPGPPGGAGAPGRPGKPGRPGAAGMPGNPGKSGGSPCDPKTPPPCKPCPAGE
PGPAGPPGPPGDAGPDGAPGNPGYGSIPGEPGPKGPPGPDGQPGYPGSPGQPGQPAYSTPPSPGEPGPQGPPGPQGPPGP
DGQPGYGGGQGQPGPKGPPGPDGQPGPDGNPGQPGYPGQPGGPGEPGICPKYCAIDGGVFFEDGTRR
```

I think that, when fetching sequences, entries (and their locations) could be ignored so each sequence would be reported once, whether an entry has multiple hits on the same protein, or multiple entries hit the same protein.

PR #391, opened 2022-06-01
script templates

gustavo-salazar:
FIX: #387 modifying the script templates to avoid duplication of sequences

Fixing python2, python3 and perl. The JS didn't include the coordinates, so it wasn't duplicating the sequences

--

Also adding a warning for JS dependencies and considering the escaped `|` in the URL


tgrego (commented):
Focused more on the perl template, no major issues just a semicolon than can be added.
Should fix the issue

src/snippets/template.pl.tmpl
```diff
@@ -139,30 +139,31 @@ sub main {
         print "\n";
       <% } else if (fileType === "fasta") { %>
       if (defined $item->{"entries"}){
+        print ">$item->{metadata}->{accession}$HEADER_SEPARATOR";
         foreach my $entry (@{$item->{"entries"}}) {
+          print " $entry->{accession}(";
```
matthiasblum, line 144: The leading space before the entry accession could be removed if there is one entry only. And if there is more than one, it could be replace with another separator.

src/snippets/template.py.tmpl
```diff
@@ -121,30 +121,37 @@ def output_list():
...
+      entries = None
+      if ("entry_subset" in item):
+        entries = item["entry_subset"]
+      elif ("entries" in item):
+        entries = item["entries"]
+      
+      if entries is not None:
+        entries_header = ""
+        for entry in entries:
+          entries_header += " " + entry["accession"] + "(" + ";".join(
```
matthiasblum, line 133: The leading space before the entry accession could be removed if there is one entry only. And if there is more than one, it could be replace with another separator.

src/snippets/template.py2.tmpl
```diff
@@ -111,29 +111,36 @@ def output_list():
...
+      entries = None
+      if ("entry_subset" in item):
+        entries = item["entry_subset"]
+      elif ("entries" in item):
+        entries = item["entries"]
+      
+      if entries is not None:
+        entries_header = ""
+        for entry in entries:
+          entries_header += " " + entry["accession"] + "(" + ";".join(
```
matthiasblum, line 123: The leading space before the entry accession could be removed if there is one entry only. And if there is more than one, it could be replace with another separator.

matthiasblum (approved):
Looks good to me. A few minor comments, but nothing important.

One thing to note is that Python 2 is not maintained anymore and its usage is strongly discouraged. Maybe we could remove the Python 2 snippet?

src/components/DownloadForm/index.js
```diff
@@ -42,7 +42,7 @@ const SchemaOrgData = loadable({
 });
 
 const extractDataFromHash = (hash) => {
-  const [path, fileType, subset] = hash.split('|');
+  const [path, fileType, subset] = hash.replace('%7C', '|').split('|');
   const output = { fileType, subset: !!subset };
   const [href, params] = path.split('?');
```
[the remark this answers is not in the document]
gustavo-salazar: I thought it wasn't necessary, but then I remember modifiers can be part of the generated URL, so you are correct, rather make sure the escape codes are included. I also added one for `&`

gustavo-salazar:
> Looks good to me. A few minor comments, but nothing important.
> 
> One thing to note is that Python 2 is not maintained anymore and its usage is strongly discouraged. Maybe we could remove the Python 2 snippet?

OK Python 2 removed

PR #395, opened 2022-06-29
exact search by gene

gustavo-salazar:
Added support to the search functionality. If the user starts a search using a gene name proteins from the key organisms will be retrieved as suggestions of an exact gene match.

It also reenables the search by entry_name.

Add some tests for `<ExactMatch />`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1691,7 +1691,7 @@
         "source-map": {
           "version": "0.5.7",
           "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
-          "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w="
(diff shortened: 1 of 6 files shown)


gustavo-salazar:
> However, there are genes matching several hundreds of proteins, such as _CYTB_. In these cases, the info box is going to be huge. Would it be possible to have the exact matches displayed in a table, like the search results from EBI Search?

I wouldn't make a table with them because it would require more data for the other columns and gives the false impression of being able to paginate. But I see your point, what about displaying up to 10 or 20 proteins per species, with a `show more` link.

matthiasblum:
It's true that additional data would have to be returned from the API (e.g. accession, identifier, name, reviewed status). And I thought the table could be paginated with all data already cached in the client.
However it may indeed involve too much work for an edge case. I guess your solution is more efficient. :+1:

matthiasblum (approved):

PR #398, opened 2022-07-14
Pfam changes

gustavo-salazar:
This includes a set of changes to be ready for the decommission of the Pfam website.

* Announcement in the home page the Pfam website is to be retired
* Links to IDA and seed alignments added in Pfam sets, while listing the entries.
* The downloads page now is split in tabs:
  * InterPro
  * InterProScan
  * Pfam **(new)**

### DIFF
--- a/src/components/Matches/index.js
+++ b/src/components/Matches/index.js
@@ -684,7 +684,7 @@ const Matches = (
               search: { type: 'seed' },
             })}
           >
-            Seed Aligment
+            Link
           </Link>
         )}
       >
--- a/src/components/Matches/index.js
+++ b/src/components/Matches/index.js
@@ -668,7 +668,7 @@ const Matches = (
         displayIf={
           primary === 'entry' &&
           secondary === 'set' &&
-          description?.entry?.db == 'pfam'
+          description?.entry?.db === 'pfam'
         }
         renderer={(accession /*: string */) => (
           <Link
@@ -688,7 +688,36 @@ const Matches = (
           </Link>
         )}
       >
-        Links
+        Aligments
+      </Column>
+      <Column
+        dataKey="accession"
+        defaultKey="ida"
+        headerClassName={f('table-center')}
+        cellClassName={f('table-center')}
+        displayIf={
+          primary === 'entry' &&
+          secondary === 'set' &&
+          description?.entry?.db === 'pfam'
+        }
+        renderer={(accession /*: string */) => (
+          <Link
+            to={(customLocation) => ({
+              description: {
+                main: { key: 'entry' },
+                entry: {
+                  db: customLocation.description.set.db,
+                  accession,
+                  detail: 'domain_architecture',
+                },
+              },
+            })}
+          >
+            IDA Table
+          </Link>
+        )}
+      >
+        Domain architectures
       </Column>
     </Table>
   );
--- a/src/components/Matches/index.js
+++ b/src/components/Matches/index.js
@@ -660,6 +660,36 @@ const Matches = (
       >
         Actions
       </Column>
+      <Column
+        dataKey="accession"
+        defaultKey="seedAlignment"
+        headerClassName={f('table-center')}
+        cellClassName={f('table-center')}
+        displayIf={
+          primary === 'entry' &&
+          secondary === 'set' &&
+          description?.entry?.db == 'pfam'
+        }
+        renderer={(accession /*: string */) => (
+          <Link
+            to={(customLocation) => ({
+              description: {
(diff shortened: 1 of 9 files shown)


src/pages/Download/InterProScan/index.js
```diff
@@ -0,0 +1,38 @@
...
+      </Link>
+      . You can find, clone, and download the full InterProScan source code on
+      the{' '}
+      <Link
+        href="//github.com/ebi-pf-team/interproscan"
+        className={f('ext')}
+        target="_blank"
+      >
+        Github repository
+      </Link>
```
matthiasblum, line 32:
```suggestion
      . InterProScan's source code is available on{' '}
      <Link
        href="//github.com/ebi-pf-team/interproscan"
        className={f('ext')}
        target="_blank"
      >
        Github
      </Link>
```

src/pages/Download/InterPro/index.js
```diff
@@ -0,0 +1,251 @@
...
+
+import ipro from 'styles/interpro-new.css';
+import ebiGlobalStyles from 'ebi-framework/css/ebi-global.css';
+import fonts from 'EBI-Icon-fonts/fonts.css';
+
+const f = foundationPartial(ebiGlobalStyles, fonts, ipro);
+
+const InterProDownloads = () => {
+  const IS_UNIPARC_READY = true;
+  const UniparcLink = IS_UNIPARC_READY ? Link : 'span';
```
matthiasblum, line 15: If `IS_UNIPARC_READY` is always `true`, maybe we don't need it.

src/components/Matches/index.js
```diff
@@ -660,6 +660,65 @@ const Matches = (
...
+                },
+              },
+              search: { type: 'seed' },
+            })}
+          >
+            Seed Aligment
+          </Link>
+        )}
+      >
+        Aligments
```
matthiasblum, line 691:
```suggestion
        Seed alignment
```

src/components/Matches/index.js
```diff
@@ -660,6 +660,65 @@ const Matches = (
...
+                entry: {
+                  db: customLocation.description.set.db,
+                  accession,
+                  detail: 'entry_alignments',
+                },
+              },
+              search: { type: 'seed' },
+            })}
+          >
+            Seed Aligment
```
matthiasblum, line 687:
```suggestion
            Link
```

src/components/Matches/index.js
```diff
@@ -660,6 +660,65 @@ const Matches = (
...
+                main: { key: 'entry' },
+                entry: {
+                  db: customLocation.description.set.db,
+                  accession,
+                  detail: 'domain_architecture',
+                },
+              },
+            })}
+          >
+            IDA Table
```
matthiasblum, line 716:
```suggestion
            Link
```

src/pages/Home/index.js
```diff
@@ -182,25 +182,16 @@ const Announcement = () => (
...
-        </Link>
-        .
+        <p>
+          The Pfam website will be decomissioned on October 5th 2022. All Pfam
+          data for this, and future versions, will be available here on the
+          InterPro website. You can read more about it in our{' '}
+          <a href="http://proteinswebteam.github.io/interpro-blog/">
+            blog post
+          </a>
+        </p>
```
matthiasblum, line 194: Wouldn't it be better to say that from October 5th, we will enter a transition phase during which the Pfam website (https://pfam.xfam.org) will redirect here, but a legacy website (not updated) will be available until January?

matthiasblum (approved):
Looks good to me. A few minor comments, but nothing big in my opinion.

src/pages/Download/InterPro/index.js
gustavo-salazar: Yeah I was trying to remember about that one. I think it was something done to be ready for a change, but I'm assuming that change has already happen and it just stayed there... I'll remove it

src/pages/Home/index.js
gustavo-salazar: I think most of that info will be on the Pfam landing/redirecting page. TBH I  would avoid motivating users to use the legacy instance as it will be a downgraded version, so we don't want to have as many users there as currently are in Pfam.

gustavo-salazar: But happy to change the wording tho, maybe something to talk with all the group?

gustavo-salazar: I'll merge with the current text, but if y'all agree on a different text, just do a direct commit.

matthiasblum (approved):
All good!

PR #401, opened 2022-07-22
Fix clan search

### LINKED ISSUE #399: Some Pfam changes do not work with text search redirections
Since #398, links to seed alignments and domain architectures are displayed when listing a Pfam clan's entries.

However, when searching for a Pfam clan accession, we are redirected to `/interpro/set/all/<accession>/`, and when switching the the "Entries" tab, the links are not visible, e.g. searching `CL0003` redirects to https://wwwdev.ebi.ac.uk/interpro/set/all/CL0003/ and the list of entries (https://wwwdev.ebi.ac.uk/interpro/set/all/CL0003/entry/all/#table) does not include the link to seed alignments and domain architectures.

matthiasblum:
Resolves #399 

Updates the description object with the source database of the clan so the URL is `/set/<database>/<accession>/` instead of `/set/all/<accession>`.

### DIFF
--- a/src/utils/processDescription/handlers/index.js
+++ b/src/utils/processDescription/handlers/index.js
@@ -537,11 +537,14 @@ export const setAccessionHandler /*: Handler */ = handlerConstructor({
     value: (value) => value,
   },
   match: {
-    value(current) {
+    value(current, description) {
       // $FlowFixMe object-this-reference
       const _current = this.cleanUp(current);
-      for (const { re } of setDBs) {
-        if (re.test(_current)) return true;
+      for (const { name, re } of setDBs) {
+        if (re.test(_current)) {
+          description.set.db = name;
+          return true;
+        }
       }
     },
   },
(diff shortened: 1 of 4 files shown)


src/utils/processDescription/handlers/index.js
```diff
@@ -537,11 +537,14 @@ export const setAccessionHandler /*: Handler */ = handlerConstructor({
...
   match: {
-    value(current) {
+    value(current, description) {
       // $FlowFixMe object-this-reference
       const _current = this.cleanUp(current);
-      for (const { re } of setDBs) {
-        if (re.test(_current)) return true;
+      for (const { name, re } of setDBs) {
+        if (re.test(_current)) {
+          description.set.db = name;
```
gustavo-salazar, line 545: This feels sort of wrong for mutability reasons. I would have to dig into the code a bit more to see how is that we are using the `match` function and see if is there any risk of side effects.

gustavo-salazar (approved):

PR #404, opened 2022-08-30
Show number of sequences

matthiasblum:
This PR updates the `<select>` element listing available alignments (Pfam and InterPro entries only) so the number of sequences is displayed:

dropdown

Requires REST API changes described in ProteinsWebTeam/interpro7-api#88.

### DIFF
--- a/src/components/BrowseTabs/BrowseTabsLink/index.js
+++ b/src/components/BrowseTabs/BrowseTabsLink/index.js
@@ -14,7 +14,7 @@ import styles from './style.css';
 const f = foundationPartial(styles);
 
 const singleEntityNames = new Map(
-  Array.from(singleEntity).map(e => [e[1].name, e[0]]),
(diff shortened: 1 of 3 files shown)


gustavo-salazar:
Happy with the changes. 

However, I think that now that the number of sequences is available in that call we don't need to do the `loadData` here:
https://github.com/ProteinsWebTeam/interpro7-client/blob/fa21f9658a6b70c269cf47f01ce594ec27ceaff3/src/subPages/EntryAlignments/index.js#L109

But that  would require changing `_Alignment` to take the `num_sequences` from a prop instead of payload

gustavo-salazar:
> Should we remove the `annotation:info` modifier from the API? I don't think it is used anywhere else.

Mhh, I'm not a fan of removing endpoints of the API as who knows if is there anyone using it in a script.

Maybe we should come up with a deprecation process for these cases, like checking the API logs for its usage after not been in use by the website, and warn users about it, etc.

PR #406, opened 2022-09-01
Multiple sequence search

gustavo-salazar:
The purpose of this PR is to support the submission of multiple sequences to the InterProScan web service. 

Several changes were required in order to get this working:
1. Updates in the checks done in the textarea for the sequence
2. Updates in the "automatic clean" function to avoid deleting sequences and prevent 2 fasta headers without content
3. The results are now displayed on a "local-first" basis. Only if there is no data in IndexedDB, check the web service.
4. Each sequence on a job receives the ID of the job concatenated with a suffix `-n` where n is the position on the results file.
5. Multiple sequences are grouped in the job list, using the same logic as nucleotide files.
6. A selector of sequences in the same job was added at the top of the results page.
7.  Importing a job by its ipscan id should work with or without the suffix

Currently deployed in http://wp-np3-aa.ebi.ac.uk/future/ for testing purposes

### DIFF
--- a/src/components/IPScan/Actions/Group/index.js
+++ b/src/components/IPScan/Actions/Group/index.js
@@ -1,0 +1,77 @@
+// @flow
+import React from 'react';
+import T from 'prop-types';
+
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+
+import Tooltip from 'components/SimpleCommonComponents/Tooltip';
+import { deleteJob } from 'actions/creators';
+
+import { foundationPartial } from 'styles/foundation';
+
+import ipro from 'styles/interpro-new.css';
+import fonts from 'EBI-Icon-fonts/fonts.css';
+import local from '../style.css';
+
+const f = foundationPartial(fonts, ipro, local);
+
+/*::
+type Props = {
+  group: string,
+  deleteJob: (string)=>void,
+  jobs: {
+    [string]: {
+      metadata: {
+        group:string
+      }
+    }
+  },
+}
+  */
+const GroupActions = ({ group, jobs, deleteJob } /*: Props */) => {
+  const handleDelete = () => {
+    // prettier-ignore
+    (Object.values(jobs)/*: any */)
+      .filter(
+        ({ metadata } /*: {metadata:{group: string}} */) =>
+          metadata.group === group,
+      )
+      .forEach((job) => deleteJob(job));
+  };
+
+  return (
+    <nav className={f('buttons')}>
+      <Tooltip
+        title={
+          <div>
+            <b>Delete Job</b>: This will remove the stored data of all the
+            sequences associated with this job from your browser. Remember that
+            search results are only retained on our servers for 7 days
+          </div>
+        }
+      >
+        <button
+          className={f('icon', 'icon-common', 'ico-neutral', 'group')}
+          onClick={handleDelete}
+          data-icon="&#xf1f8;"
+          aria-label="Delete Results"
+        />
+      </Tooltip>
+    </nav>
+  );
+};
+GroupActions.propTypes = {
+  group: T.string.isRequired,
+  jobs: T.object,
+  deleteJob: T.func,
+};
+
+const mapStateToProps = createSelector(
+  (state) => state.jobs,
+  (jobs) => ({ jobs }),
+);
+
+export default connect(mapStateToProps, {
+  deleteJob,
+})(GroupActions);
--- a/src/components/IPScan/Actions/__snapshots__/test.js.snap
+++ b/src/components/IPScan/Actions/__snapshots__/test.js.snap
@@ -8,7 +8,7 @@ exports[`<Actions /> should render 1`] = `
     title={
       <div>
         <b>
-          Delete search
+          Delete sequence
         </b>
         : This will remove the stored data from your browser. Remember that search results are only retained on our servers for 7 days
       </div>
--- a/src/components/IPScan/Actions/index.js
+++ b/src/components/IPScan/Actions/index.js
@@ -105,7 +105,7 @@ export class Actions extends PureComponent /*:: <Props> */ {
         <Tooltip
           title={
             <div>
-              <b>Delete search</b>: This will remove the stored data from your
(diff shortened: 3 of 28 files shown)


matthiasblum:
I haven't looked at the code yet, but I have a few comments. They are not necessarily regarding this PR, but broadly concern how we display InterProScan search results :monocle_face: 

#### Job submission

Setting a job title renames the name of the first sequence. This is fine when submitting a single sequence, but not really helpful when submitting more.



When turning on *Create another job after this one*, there are no confirmation the job is submitted: the text are is cleared and a *Automatic FASTA clean up* button is displayed:



I am not sure what it does, but it disappears when pasting new sequences in the text area.

#### Running jobs

The job name is the name of the first sequence. That seems counter-intuitive.



#### Jobs list

We show the group (job itself) and the individual sequences. This becomes an issue when submitting many sequences, as we need to navigate several pages and there are no search feature (i.e. searching sequence by name).

We could display the _job_ only (name, date, number of sequences, status, delete button, download button).



I don't think the job ID (e.g. `iprscan5-R20220902-112341-0506-55207823-np2`) is very important in most cases. It can be displayed on the job page itself, but in the results page, we could show a more informative name if we can:

- the job title, if defined by the user
- the sequence name, if only one sequence was submitted
- the file name, if sequences come from a file instead of a copy-paste
- if none of the above: `<year><month><day>-<hour><minute><second>-<number_of_sequences>`

####  Job result



1. The navigation should be below the page title
2. The title (the sequence name) should not be editable
3. It would be nice to use the sequence name as the sequence job ID (currently job ID + suffix)
4. As the data is stored in IndexDB as soon it's available, it is not necessary to have a _Save results in browser_ button. I also think we should not be able to delete the results of a single sequence. Only the entire job should be deletable.
5. Same as above, show an expiration date could lead users to think their results (which are stored in their browser) will be deleted on the date shown.

---

The way I see it, the jobs list page (`result/InterProScan`) should only display InterProScan jobs, not individual sequences. The job page (`result/InterProScan/<job-name>/`) should list the sequences (maybe in a searchable table), with the number of reported matches (if the job is complete), and a result page (`result/InterProScan/<job-name>/<sequence-name>/`) should show the matches.

gustavo-salazar:
Thanks a lot for all the detailed feedback. Here are some inline comments:

> #### Job submission
> Setting a job title renames the name of the first sequence. This is fine when submitting a single sequence, but not really helpful when submitting more.
1. up until now there was no distinction between sequence header and job title, because it was a 1 to 1 relationship. This title/name/id was used mostly for the user to be able to have something memorable for their sequence, and if change it it was immediately used as a sequence header. I agree is worth separating those concepts, I will think if there is any easy way of achieving this without a major refactoring of this area.

> When turning on _Create another job after this one_, there are no confirmation the job is submitted: the text are is cleared and a _Automatic FASTA clean up_ button is displayed:
> I am not sure what it does, but it disappears when pasting new sequences in the text area.
2. OK I'll add a confirmation message and make sure the clean-up button won't get displayed unnecessarily.

> 
> #### Running jobs
> The job name is the name of the first sequence. That seems counter-intuitive.

See [1]

> 
> 
> 
> #### Jobs list
> We show the group (job itself) and the individual sequences. This becomes an issue when submitting many sequences, as we need to navigate several pages and there are no search feature (i.e. searching sequence by name).

3. The way this is displayed was "stolen" from the cases when uploading a nucleotide file, in such cases, the group name was the sequence name, and each ORF got its own sequence. So, this was a very quick solution to develop as literally was only a matter of adding a value to `metadata.group`. But of course, a solution for up to 6 sequences is not 100% sufficient to the now up to 100.
4. Not what you are asking, but a search by sequence name should be quite simple to add

> 
> We could display the _job_ only (name, date, number of sequences, status, delete button, download button).
> 
5. Yes I think this is possible as part of a bit of refactoring this part as you suggest at the end.
> 
> I don't think the job ID (e.g. `iprscan5-R20220902-112341-0506-55207823-np2`) is very important in most cases. It can be displayed on the job page itself, but in the results page, we could show a more informative name if we can:
6.  I think the ID it is actually quite useful as it is the only recognisable ID out of the website/browser. You could share this ID and another user recovers the job in their browser, or get other results formats from the www-prod web service. Nonetheless I see how visually is not that important, and could get a bit less exposure.
> 
> * the job title, if defined by the user
> * the sequence name, if only one sequence was submitted
> * the file name, if sequences come from a file instead of a copy-paste
> * if none of the above: `<year><month><day>-<hour><minute><second>-<number_of_sequences>`
7. cool, sensible way to get the Job title.
> 
> #### Job result
> 
> 1. The navigation should be below the page title

OK

> 2. The title (the sequence name) should not be editable

unsure of this one. I think there might be cases (e.g. a typo) when you might want to change the title of one sequence.

> 3. It would be nice to use the sequence name as the sequence job ID (currently job ID + suffix)

A lot of the logic is based on the IDs having this format, so changing this would actually imply major changes.

> 4. As the data is stored in IndexDB as soon it's available, it is not necessary to have a _Save results in browser_ button. I also think we should not be able to delete the results of a single sequence. Only the entire job should be deletable.

We are still deleting it from IndexedDB after 7 days unless the user has pressed the save button. But we can change this to just keep it "saved" from the beginning

> 5. Same as above, show an expiration date could lead users to think their results (which are stored in their browser) will be deleted on the date shown.

See above

> 
> The way I see it, the jobs list page (`result/InterProScan`) should only display InterProScan jobs, not individual sequences. The job page (`result/InterProScan/<job-name>/`) should list the sequences (maybe in a searchable table), with the number of reported matches (if the job is complete), and a result page (`result/InterProScan/<job-name>/<sequence-name>/`) should show the matches.

This seems sensible but will require a bit of work to change it to it, unsure if including this in this PR or opening another story for it

matthiasblum (approved):
Happy with the changes.

A few things could be improved, such as not having an empty job title just after submitting sequences, or being able to download all results from one job.

I still think that it would be clearer to have a page with the list of job, a page for a given job (where sequences are listed), and a page for the results of a given sequence. Maybe for another time. :innocent:

PR #407, opened 2022-09-02
Subfamilies

gustavo-salazar:
Adding a subpage for panther families to list the subfamilies related to it.

It also adds a section in the popups in the protein viewers where a panther family matches, indicating which subfamily is responsible for the match.

A small improvement in the way the protvista popups work was made to avoid the creation of multiple react roots and instead reuse the first one that was created.

**Note:** It requires the API change in https://github.com/ProteinsWebTeam/interpro7-api/pull/89

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -45,6 +45,7 @@ pages:
       - pathways
       - genome3d
       - curation
+      - subfamilies
   protein:
     subPages:
       - entry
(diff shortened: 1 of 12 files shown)


matthiasblum (changes requested):
Thank you. There are few things to change, but it shouldn't be long/difficult.

However, an error is raised on `/interpro/entry/cathgene3d/G3DSA:1.10.10.10/subfamilies/` with the following message:
> "G3DSA:1.10.10.10" is not a valid value for "entry.accession"

Do you manage to reproduce it?

PR #409, opened 2022-09-16
Relayout protvista

gustavo-salazar:
Changing the layout of the protvista component for a grid layout to make it more responsive, and to fix some bugs on the options to `Save as image` or `print`.

I have merged this in `future` so it should auto-deploy in http://wp-np3-aa.ebi.ac.uk/future/ tomorrow

If you find an issue with one of the export options please let me know which browser and the size of the window.


typhainepl:
Looking better, however I now have a grey lines on the identifiers column. Imaged attached is through the "save as image" option, but it is displayed the same way on the web page.
protvista (2)
Also, do you think it would be possible to have the printed version on a single page? At the moment it prints on multiple pages.

src/components/ProtVista/index.js
```diff
@@ -609,282 +621,253 @@ export class ProtVista extends Component /*:: <Props, State> */ {
           <div className={f('popper-content')} ref={this._popperContentRef} />
```
matthiasblum, line 621: That might seem like a stupid comment, but I could not find where the `.popper-content` rule is defined. Do we actually need to this class?

matthiasblum (approved):
Looks good. 
On the current website, the view overflows for some proteins (e.g. https://www.ebi.ac.uk/interpro/protein/reviewed/P43403/) and this PR fixes this: brilliant!

gustavo-salazar:
> Looking better, however I now have a grey lines on the identifiers column. Imaged attached is through the "save as image" option, but it is displayed the same way on the web page.

Ok, I found a way to deal with that issue, now the background of the scrollbars is transparent, so even when they are always visible, they won't affect the style in there. I will let you know when I deploy this change.

> Also, do you think it would be possible to have the printed version on a single page? At the moment it prints on multiple pages.

That one is a bit more complicated, because I'm forcing a width at the moment of printing to make sure the layout is respected, so I dunno how to scale that width for the size of the page, I think there are too many variables in there to take into account to get it right. Unsure if it is possible or even worth the effort.

src/components/ProtVista/index.js
gustavo-salazar: yeah , I checked and it's not defined, it might have been used before and removed the rule 🤷🏽
I'll remove the class from that element.

PR #412, opened 2022-09-27
Release91 struct models

matthiasblum:
This PR accommodates two data changes for the upcoming release:

- the counter for RoseTTAFold structures has been renamed from `RoseTTAFold` to `rosettafold`
- the number of proteins with an AlphaFold model is calculated for all entries, and not only InterPro/Pfam: 51c075f3c7ba039f566e2b5ba4f7e614bf6c1bd6 adds a condition to only display the menu link for InterPro/Pfam entries


gustavo-salazar (approved):
Happy with the changes. Just curious, if the number of AlphaFold models has been calculated for all entries, why are we restricting that page for only Pfam and InterPro

matthiasblum:
It's true we could simply show the AlphaFold tab for all member databases. We can have a quick talk with Alex tomorrow.

gustavo-salazar (approved):

PR #417, opened 2022-10-12
Feedback pfam

gustavo-salazar:
3 minor fixes:

* IDA results -width scalable to the longest protein in the page
* Short name in the list of entries of a clan
* Adding fragment tag to proteins


matthiasblum (approved):
Looks good to me.

PR #418, opened 2022-10-14
Fix download link of Pfam alignments

matthiasblum:
`EntryAlignments` takes five arguments: `data`, `customLocation`, `type`, `url`, `showCtrlToZoomToast`, and `goToCustomLocation`.

In 75e5304369ea825a8d4eec13ae9a37ab1919cf7d, I stupidly removed the third argument passed to `createSelector` :man_facepalming:: when creating an instance of `EntryAlignments`, the value for `url` (string) became the value of `showCtrlToZoomToast` (boolean) causing the download link to be something like `truealignment:full&download`.

This PR fixes that bug.


gustavo-salazar (approved):
OK, happy with the change... I'll merge and after the CI passes on `dev` I'll merge into `master`

PR #421, opened 2022-10-24
confidence track

gustavo-salazar:
Alphafold provides a JSON that has the lPDDT score per amino acid, for example for the [Q8I3H7 model](https://alphafold.ebi.ac.uk/files/AF-Q8I3H7-F1-confidence_v3.json). 

We are now retrieving this file when displaying the protein viewer at the side of an alphafold model in order to generate a track that shows this information in the context of the protein sequence.

This PR uses the nightingale coloured track, defining the sequence for it as the categories retrieved on the file (`"D" |  "L" | "M" | "H"`) and defining the attributes of the track as:

```html
<protvista-coloured-sequence
  scale="H:90,M:70,L:50,D:0"
  color_range="#ff7d45:0,#ffdb13:50,#65cbf3:70,#0053d6:90,#0053d6:100"

  ...

/>
```

Other changes included in this PR:
* The Color Scale in the PopUps now support scales with multiple points as the one in the AlphaFold track
* Adjusting the style of the legends on the Scale popups
* The `width` of the title on the popups is now limited to `20ch` to avoid popups that take up the whole screen.
* Adding label for hydrophobicity track

### DIFF
--- a/src/subPages/AlphaFoldModelSubPage/index.js
+++ b/src/subPages/AlphaFoldModelSubPage/index.js
@@ -25,6 +25,7 @@ const AlphaFoldModelSubPage = ({ data, description }) => {
   const [selectionsInModel, setSelectionsInModel] = useState(null);
   const [proteinAcc, setProteinAcc] = useState('');
   const [modelId, setModelId] = useState(null);
+  const [confidenceURL, setConfidenceURL] = useState(null);
   const handleProteinChange = (value) => {
     setProteinAcc(value);
     setModelId(null);
@@ -54,6 +55,12 @@ const AlphaFoldModelSubPage = ({ data, description }) => {
           onModelChange={handleModelChange}
           modelId={modelId}
           selections={selectionsInModel}
+          onModelCifChange={(cif) => {
+            const newURL = cif?.length
+              ? cif.replace('-model', '-confidence').replace('.cif', '.json')
+              : null;
+            if (newURL !== confidenceURL) setConfidenceURL(newURL);
+          }}
         />
       )}
       {mainType === 'entry' ? (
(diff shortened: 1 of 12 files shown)


src/subPages/AlphaFoldModelSubPage/index.js
```diff
@@ -54,6 +55,12 @@ const AlphaFoldModelSubPage = ({ data, description }) => {
           onModelChange={handleModelChange}
           modelId={modelId}
           selections={selectionsInModel}
+          onModelCifChange={(cif) => {
+            const newURL = cif?.length
+              ? cif.replace('-model', '-confidence').replace('.cif', '.json')
+              : null;
+            if (newURL !== confidenceURL) setConfidenceURL(newURL);
+          }}
```
matthiasblum, line 63: Wouldn't it be possible to handle this in `handleProteinChange()`?

matthiasblum (approved):
Looks good to me. 

I got slightly confused with `onModelCifChange`, so I added a comment, but even if it wasn't necessary to add a new parameter to `AlphaFoldModel`, I'm happy with the changes as they are.

src/subPages/AlphaFoldModelSubPage/index.js
gustavo-salazar: nope, `onModelChange` gets trigger when the Select has changed but at that point, the request to get the metadata with links from alphafold is only to be started, and I need the Cif from the response, so basically this method is about knowing the first alphafold request has been received

gustavo-salazar:
> Looks good to me.
> 
> I got slightly confused with `onModelCifChange`, so I added a comment, but even if it wasn't necessary to add a new parameter to `AlphaFoldModel`, I'm happy with the changes as they are.

I don't see the comment you added. Maybe you forgot to push?

PR #423, opened 2022-10-28
GO terms for Subfamilies

gustavo-salazar:
This PR lists the GO terms provided by panther in 2 parts:
1. In the protein page if the protein has a match with a panther family, the client retrieves the information of the subfamily matching, including its Go terms and displays them under the list of InterPro GO terms. e.g. `/protein/uniprot/A0A010R210/`
2. In the subfamilies subpage, as an additional column on the list of subfamilies. It only displays the first 10 GO terms, with the option to display more on demand. e.g. `/entry/panther/PTHR10026/subfamilies`

This requires the changes on the API that are part of https://github.com/ProteinsWebTeam/interpro7-api/pull/92

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,23 +24,23 @@
       }
     },
     "@babel/code-frame": {
-      "version": "7.16.7",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.7.tgz",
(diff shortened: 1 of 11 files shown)


matthiasblum (approved):
Looks good. A small improvement would be to display the name of GO terms in the subfamilies table instead of their ID. I saw the category and name are displayed in a tooltip when hovering the ID, but users may want to search for GO terms using key words, e.g. _regulation_, _binding_, etc.

gustavo-salazar:
Yeah, I initially have put the name but considering space in the page I went for just the ID. But later we added the `Read More` button so I guess I can go back to display the name, or do you think I should display both name and ID?

PR #424, opened 2022-10-28
IDA - support to display names instead of accession

gustavo-salazar:
Gives the option to display entry names instead of accessions in the IDA search results.

It also has a bit of a refactoring of the DomainArchitecture components, just splitting them into a file per component

**Note 1** Don't merge until new data release. This requires changes in the elasticsearch index and won't work until the names are included in the `ida` index.

**Note 2** It merges into `subfamilies` because that's from where I branch out 🤦🏽‍♂️, but we should change it to merge into `dev` once #423 is merged


matthiasblum (approved):
Looks good.

However, is there a reason for letting users choose whether the label is the accession or the name? Since accessions are always displayed above the domain architecture, I would ditch the switch and display names.

gustavo-salazar:
Fair enough. 
I'm gonna hide that switch and default the redux value to `name`, just in case we later think it was useful.

But I will put a comment in it to be removed completely if in 6 months we haven't changed our minds.

PR #426, opened 2022-11-01
Bugfix pfam-n + unintegrated pfam

### LINKED ISSUE #425: Missing Pfam domain display in sequence viewer for Q2LD37
The Pfam domain isn't displayed in the unintegrated category of the sequence viewer, only the ID on the right hand side is displayed.
[Q2LD37](https://www.ebi.ac.uk/interpro/protein/UniProt/Q2LD37/)
image

gustavo-salazar:
FIX #425 
it concatenates `-n` on accessions that come from Pfam-N to avoid using the same ID as an Unintegrated pfam track

### DIFF
--- a/src/components/ProtVista/index.js
+++ b/src/components/ProtVista/index.js
@@ -99,6 +99,9 @@ const loadProtVistaWebComponents = () => {
   return Promise.all(webComponents);
 };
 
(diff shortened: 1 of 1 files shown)


matthiasblum (approved):
Looks good to me. :+1:

PR #427, opened 2022-11-02
IPScan batch: group actions  delete and rerun

gustavo-salazar:
In this PR 2 new actions are added to Ipscan jobs that have been run in batch:
* Download All
* Re-run all jobs

These options have also been grouped into a dropdown menu.

It also fixes some issues related to the re run of jobs with a subset of the memeber DBs

### DIFF
--- a/src/components/IPScan/Actions/Group/index.js
+++ b/src/components/IPScan/Actions/Group/index.js
@@ -6,7 +6,9 @@ import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
 
 import Tooltip from 'components/SimpleCommonComponents/Tooltip';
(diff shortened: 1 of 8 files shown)


matthiasblum (commented):
Looks good.

A few suggestions, if possible:

- remove icon in the "Group actions" dropdown. I don't think they are necessary to illustrate the action, and they are not always on the same line as the text:
image
Alternatively, you can apply `white-space:nowrap;` to each button to fix this.
- is it possible to download all results in other formats than JSON?
- "Clear all" options: make it a real dropdown. It is currently two buttons separated by an `<hr>`. Also, the button could have `cursor: pointer` applied to them.

PR #436, opened 2022-11-24
Confidence track in the protein page

gustavo-salazar:
Adds the AlphaFold confidence track in the protein viewer of the protein page
Removes the hydrophobicity track.

Refactor the way we get the data for the confidence track so it can be reused.


matthiasblum (approved):
Looks good to me.

PR #440, opened 2022-12-06
Add .gz suffix to download file name

matthiasblum:
When downloading the HMM model, the extension is `.hmm` but the file is gzip-compressed, which puzzled some users. This PR adds the `.gz` suffix.


gustavo-salazar (approved):
All good

PR #445, opened 2023-01-06
using the set info from the entry payload

gustavo-salazar:
Now the entry payload returns a JSON in `set_info` that contains the `name` and `accession` of the set that an entry belongs to. null, if it doesn't belong to any set.

This PR uses that info to avoid the extra request to display the set on the entry page.


matthiasblum (approved):
lgtm :+1:

PR #446, opened 2023-01-06
Styling the popups in the protein viewer

gustavo-salazar:
This PR makes some style changes in the popups that are part of the protein viewer.

Among other things trying to reduce the size of them and that way be less intrusive while inspecting the entry matches. 

Similar changes are included for the Domain architecture results.


src/components/ProtVista/Popup/Residue/index.js
```diff
@@ -29,25 +28,20 @@ const ProtVistaResiduePopup = ({ detail, sourceDatabase } /*: Props */) => {
   return (
     <section>
       <h6>
+        Residue in {sourceDatabase} -{' '}
         {accession.startsWith('residue:')
           ? accession.split('residue:')[1]
           : accession}
-        {description && <p>[{description}]</p>}
       </h6>
```
matthiasblum, line 35: I would remove the "Residue in ". Just above the track, there is the "Residues" title.

src/components/ProtVista/Popup/Residue/index.js
```diff
@@ -29,25 +28,20 @@ const ProtVistaResiduePopup = ({ detail, sourceDatabase } /*: Props */) => {
...
-          </li>
-          <li>Residue: {residue}</li>
-        </ul>
-      </div>
+      <ul>
+        <li>
+          Position: {start}
+          {end && end !== start ? <>-{end}</> : null}
+        </li>
+        <li>Residue: {residue}</li>
```
matthiasblum, line 43: Is `end !== start` possible? The only case where I believe it could happen is if there are several residue annotations in a row, but in this case, would `residue` be a string of all the residues annotated?

Also, maybe the position and the residue code could be on the same line (not in two different list items)?

src/components/SimpleCommonComponents/DynamicTooltip/index.js
```diff
@@ -31,27 +32,25 @@ const _DataProvider = (
...
       const message = () => (
-        <>
-          <b>{accession}</b>
-          <br />
-          {name} <br />
-          <small>{databases[db].name}</small>
+        <section className={f('entry-popup')}>
+          <h6>
+            {databases[db].name} - {accession}
+          </h6>
```
matthiasblum, line 38: Is this tooltip used for anything besides domain architectures? If no, maybe it is not necessary to display `databases[db].name`.

matthiasblum (approved):
Looks good to me. Made a few comments, but nothing blocking.

src/components/SimpleCommonComponents/DynamicTooltip/index.js
gustavo-salazar: Mhh... the user can still choose between InterPro or Pfam, so would leave it there

PR #447, opened 2023-01-06
adding abutton to toggle the PiP functionality

gustavo-salazar:
The Picture in Picture strategy to display the structure viewer when the page is scrolled down was reported as obtrusive.

This PR adds a button to disable that functionality. The setting is saved in the redux state, and therefore it persists over the session.


matthiasblum (approved):
I was never a big fan of the PiP feature, so I welcome this PR with open arms! :partying_face: 

One comment (but not a blocker): when hovering controls in the AlphaFold page (e.g. `/interpro/protein/reviewed/P02185/alphafold/`), the cursor turns into a pointer, but it doesn't in the structure page (e.g. `/interpro/structure/PDB/101m/`). If possible, could you make the cursor a pointer there as well?

PR #448, opened 2023-01-06
Adding split screen functionality to the alphafold viewer

### LINKED ISSUE #444: View AlphaFold structure side by side with protein viewer
It would be nice to be able to visualise the AlphaFold structure side by side with the protein sequence viewer, like the feature available for PDB structures.

gustavo-salazar:
FIX #444 
It is useful to see the structure viewer side by side with the protein feature viewer. One way to do it is to fullscreen them and split the screen 50/50. This functionality was already available when displaying PDB structures, but it was missing for AlphaFold models, hence this PR.

**Note:** Review after #447 as this branched out of it


src/components/AlphaFold/Model/index.js
```diff
@@ -93,21 +98,23 @@ const AlphaFoldModel = (
       : models.filter((x) => x.entryId === modelId);
   const elementId = 'new-structure-model-viewer';
   return (
-    <div>
+    <div className={f('alphafold-model')}>
       <h3>
         AlphaFold structure prediction
         {models.length > 1 || hasMultipleProteins ? 's' : ''}
       </h3>
```
matthiasblum, line 105: Maybe hide this in slit-screen mode?

matthiasblum (changes requested):
Looks nice, thanks.

In full screen mode, the control buttons (PDB/mmCIF download buttons, reset image button, etc.) are hidden. However, in split full screen mode that are displayed... including the "view the structure in full screen mode", although we are already technically in a full screen mode. And clicking this button just hide the protein viewer.

I think it'd be better if the two screen modes had the same behavior with respect to the control buttons.

gustavo-salazar:
OK, changes done, please check again.

matthiasblum:
Thank you. Looks good. :+1:

matthiasblum (approved):
Looks good. Thank you.

PR #450, opened 2023-01-10
Conservation track errors

### LINKED ISSUE #384: Fetch conservation sometime fails but missing any useful message
I've noticed that sometimes the conservation API call some times returns a failure with content like `detail="results"` however the website doesn't shows any message and the panel for conservation just disappears.

Related is that when the conservation call takes too long, the `408` response, behaves correctly in the background. Meaning it gets asked again periodically until getting a result. However the conservation panel also disappears in the mean time, and the reappears when the request is finally completed.

gustavo-salazar:
FIX #384 

Adding an error message when the match conservation payload fails.
It will include the HTTP error code if available.


src/components/ProtVista/index.js
```diff
@@ -929,16 +931,24 @@ export class ProtVista extends Component /*:: <Props, State> */ {
...
+                        <>
+                          {this.state.showLoading ? (
+                            <div
+                              className={f('loading-spinner')}
+                              style={{ margin: '10px auto' }}
+                            >
+                              <div />
+                              <div />
+                              <div />
+                            </div>
```
matthiasblum, line 948: Is it not possible to import and use `Loading` from `components/SimpleCommonComponents/Loading`?

src/components/Related/DomainsOnProtein/index.js
```diff
@@ -432,25 +437,28 @@ export class DomainOnProteinWithoutMergedData extends PureComponent /*:: <Props>
...
-    ) {
-      handleLoaded(dataConservation.payload);
+    if (dataConservation && !dataConservation.loading) {
+      if (dataConservation.ok && dataConservation.payload) {
+        handleLoaded(dataConservation.payload);
+      } else {
+        handleError(dataConservation);
+      }
     }
+    console.log('=>', dataConservation);
```
matthiasblum, line 453: Remove call to `console.log`.

matthiasblum (approved):
Looks good. Too minor comments, but nothing preventing merging.

But why the conservation track is only available for reviewed proteins with PANTHER matches? And with HMMER being not so stable, wouldn't it be possible to pre-calculate all conservation tracks?

gustavo-salazar:
The first version of this feature used Pfam Families but it got changed to use panther ones because they should bring more coverage. TBH I can't remember why it runs only on a single member DB. I'm guessing we could pre-calculate this as a release procedure, it currently just runs on swissprot proteins with a panther match.

PR #458, opened 2023-02-14
Nightingale v4

gustavo-salazar:
All the changes required to use Nightingale components v4. The components/pages were the changes are relevant:
- Any protein viewer(e.g. protein, structure and model, interproscan)
- Sunburst view of the taxonomy
- Pfam alignments subpage
- Rossetta model pages have the multiple components interacting(heatmap, contacts, msa, navigation, and molstar) 

Changes include:

 - installing the new components from @nightingale-elements namespace in npm.
 - All attributes are now `snake-case`
 - `d3` was removed as a general dependency and only kept `d3-scale` because of its use of mapping coordinates between structure, protein and entries.
 - Most files changed correspond to updates on `flow-typed` sorry for the noise :-)
 - dependencies updates.

**NOTE:** I'll move this from draft to an actual PR when all the PRs in nightingale get resolved. and components are published on full version and not `alpha`


matthiasblum (changes requested):
Great work, thanks!

Only one thing that I think needs to be changed: a reference to `protvista-colored-sequence`

https://github.com/ProteinsWebTeam/interpro7-client/blob/1f1107ae744025fa40bf62b21877bde1a7e08b21/src/components/ProtVista/Popup/Hydro/index.js#L10

I saw that we depend on `draft-js` for the text box used to search sequences. Would it be possible to use Nightingale's `textarea-sequence` instead in the future?

gustavo-salazar:
> I saw that we depend on `draft-js` for the text box used to search sequences. Would it be possible to use Nightingale's `textarea-sequence` instead in the future?

Yes, thatwas the motivation to work on migrating `textarea-sequnce` at this stage. Draft.js has been deprecated, while Quill.js is still on active development

PR #459, opened 2023-02-14
History5

gustavo-salazar:
We use the [history](https://www.npmjs.com/package/history) library to keep in sync the browser history and the custom location we manage in the redux state.

In version 5 of this library, they removed the option `basename` of their config, which we were using, to for example ignore the `/interpro/` from a URL such as https://www.ebi.ac.uk/interpro/protein/UniProt/P05067/.
 
There is a github issue about it, but the developers state that there are no plans to bring `basename` back: https://github.com/ReactTraining/history/issues/810 

This PR creates a wrapper over the library to keep the `basename` in the object we use when handling the URL and being able to add/remove when mapping the URL to the state or vice-versa.

**NOTE:** This was done after #458 so only merge once that one is merged
**NOTE:** Merged to `remove-legacy-bundle` so review this after to #471


matthiasblum (approved):
Looks good to me. :+1: 

Maybe you could replace occurrences of `history.history` and `history.basename` by `historyWrapper.history` and `historyWrapper.basename` to avoid confusion in the future, but that's a small detail.

Also, after seeing this:

https://github.com/ProteinsWebTeam/interpro7-client/blob/d6b0e534efc8a101746314a42a1c8f7eff2cd1c0/src/App.js#L17

I had a look at the [removeLastSlash()](https://github.com/ProteinsWebTeam/interpro7-client/blob/d6b0e534efc8a101746314a42a1c8f7eff2cd1c0/src/utils/url/index.js#L7) function. Maybe the patterns used to trim leading/trailing slashes could be changed from

https://github.com/ProteinsWebTeam/interpro7-client/blob/d6b0e534efc8a101746314a42a1c8f7eff2cd1c0/src/utils/url/index.js#L4-L5

to

```js
const FINAL_SLASH = /\/+$/;
const START_SLASH = /^\/+/;
```

PR #468, opened 2023-03-17
organizing the tracks in the ipscan results

gustavo-salazar:
The InterProScan results include a protein viewer, this one however was displaying some categories in a different way than the viewer in the protein pages.

Some of the inconsistencies were:
* The category called Predictions should be Other features
* Funfam where displayed as Unintegrated features
* PIRSR residues were displayed as features.
* Residues were displayed together with their associated entries. However, in the protein viewers, those are displayed in a separate category.

### DIFF
--- a/src/components/IPScan/Summary/serializers.js
+++ b/src/components/IPScan/Summary/serializers.js
@@ -24,8 +24,9 @@ const integrateSignature = (signature, interpro, integrated) => {
   const accession = interpro.accession;
   const entry = integrated.get(accession) || {
     accession,
-    name: interpro.name,
-    source_database: 'InterPro',
+    name: interpro.description,
+    short_name: interpro.name,
+    source_database: 'interpro',
     _children: {},
     children: [],
     type: interpro.type.toLowerCase(),
@@ -42,7 +43,9 @@ const match2residues = (match) => {
         ? {
             accession: match?.signature?.accession || match?.accession,
             locations: sites.map((site) => ({
-              description: site.description,
+              description: `${site.label || ''}${site.label ? ': ' : ''}${
+                site.description
+              }`,
               fragments: site.siteLocations,
             })),
             type: 'residue',
@@ -117,10 +120,11 @@ const condenseLocations = (
   }));
 };
 
+// eslint-disable-next-line max-statements
 export const mergeData = (matches, sequenceLength) => {
   const mergedData /*: {
-    unintegrated:any[], 
-    other_features: any[], 
+    unintegrated:any[],
+    other_features: any[],
     residues: any[],
     other_residues: any[],
   } */ = {
@@ -130,8 +134,7 @@ export const mergeData = (matches, sequenceLength) => {
     other_residues: [],
   };
   const unintegrated = {};
-  const predictions = {};
-  const residuesCategory = [];
+  const otherFeatures = {};
   let integrated = new Map();
   const signatures = new Map();
   for (const match of matches) {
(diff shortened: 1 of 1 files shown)


src/components/IPScan/Summary/serializers.js
```diff
@@ -12,6 +12,9 @@ type Location = {
 }
 */
 
+const OTHER_FEATURES_DBS = ['funfam'];
```
typhainepl, line 15: Does InterProScan reports Pfam-N matches? If so I think Pfam-N should also be added to that list.

src/components/IPScan/Summary/serializers.js
```diff
@@ -95,12 +118,20 @@ const condenseLocations = (
...
+    residues: any[],
+    other_residues: any[],
+  } */ = {
+    unintegrated: [],
+    other_features: [],
+    residues: [],
+    other_residues: [],
+  };
   const unintegrated = {};
+  const predictions = {};
```
typhainepl, line 133: I would suggest to change `predictions` to `other_features` for consistency.

src/components/IPScan/Summary/serializers.js
```diff
@@ -121,26 +152,28 @@ export const mergeData = (matches, sequenceLength) => {
...
+        ...processedMatch,
+        accession: `residue:${processedMatch.accession}`,
+        residues,
+      });
+    }
     if (NOT_MEMBER_DBS.has(library)) {
-      processedMatch.accession += ` (${mergedData.predictions.length + 1})`;
       processedMatch.source_database = library; // Making sure the change matches the ignore list.
-      mergedData.predictions.push(processedMatch);
+      if (processedMatch.accession in predictions) {
```
typhainepl, line 170: same comment as above, regarding the `predictions` variable name

typhainepl (commented):
The changes looks good overall! A few comments below, I've been searching for the UniProt [P00519](https://www.ebi.ac.uk/interpro/protein/UniProt/P00519/) sequence.

Regarding PIRSR, in the protein viewer of the protein page, the category is shown (e.g. ACT_SITE: Proton acceptor. NP_BIND: ATP. BINDING: ATP.), would it be possible to do the same for the InterProScan search results?

In the protein viewer of the protein page, a logo is shown before the InterPro entry giving information about the type, could you do the same here?

gustavo-salazar:
OK changes done!

typhainepl:
Looking good to me.
Once Matthias has confirmed the info about Pfam-N, you can merge.

src/components/IPScan/Summary/serializers.js
matthiasblum: Correct. Pfam-N matches were provided by Google Research and are not reported by InterProScan.

src/components/IPScan/Summary/serializers.js
```diff
@@ -121,26 +155,28 @@ export const mergeData = (matches, sequenceLength) => {
...
-              accession: match.signature.accession,
-              locations: sites.map((site) => ({
-                description: site.description,
-                fragments: site.siteLocations,
-              })),
-            }
-          : null,
-      )
-      .filter(Boolean);
-    if (residues.length > 0) processedMatch.residues = residues;
```
laisecf, line 138: A doubt ^^ what did that line do exactly? Why this is not necessary anymore after changes?

laisecf (approved):

src/components/IPScan/Summary/serializers.js
gustavo-salazar: The changes were calculated in the variable `residues`, however at the end what we return is the variable `processedMatch` so the line is just putting the calculation in the output variable when it found any residues.
This, in the new code, has been changed to line 165, where the residues are now included together with `processedMatch` to get the effect that the residues are part of the entry

laisecf: ok!! thanks 🤗

PR #471, opened 2023-03-31
removing legacy bundle

gustavo-salazar:
Our webpack configuration was set up to create 2 bundles: 1 for modern browsers and 1 for legacy browsers that don't support es-modules.

Checking the stats of the last 7 days from the [livelogs](http://ves-ebi-2e.ebi.ac.uk/es/livelogs/#dashboard/temp/AYc32qFSLmmAsU_n8B8d), from 2356060 hits requesting `*.js` files from the interpro machines, only  4821 hits were requesting `*legacy*` files; for a total 0.2%.

This PR removes the `legacy` bundle, and adds warning messages to upgrade browsers for the affected cases. 

The particular change pushing in this direction is that Lit elements are highly dependent on classes, so adapting Nightingale components would require effort tuning the already complex webpack/babel/typescript setup. An effort that is hard to justify for such a small amount of requests.

**NOTE:** Probably better to review this one before #458 to be able to visualise the new components.

### DIFF
--- a/src/index.template.html
+++ b/src/index.template.html
@@ -154,15 +154,6 @@
         }
       </style>
     </noscript>
-    <% if (!htmlWebpackPlugin.options.inject) { %> 
-      <%= htmlWebpackPlugin.faviconTags %> 
-    <% } %> 
-    <% for (const css of htmlWebpackPlugin.files.css) { 
-         if (css.indexOf(htmlWebpackPlugin.entry)!==-1 || (css.indexOf('.hydrate.')===-1 && css.indexOf('.index.')===-1)) { %>
-           <link rel="stylesheet" href="<%= css %>" />
-    <%   } 
-       }
-    %>
   </head>
   <body class="ebi-black-bar-loaded">
     <noscript>
@@ -175,6 +166,7 @@
       <div class="tmp-layout">
         <div class="tmp-ebi-header"></div>
         <div class="tmp-interpro-header"><h1>InterPro</h1></div>
+        <div id="error"></div>
         <p>Loading the InterPro website, please wait.</p>
         <div class="loading">
           <div></div>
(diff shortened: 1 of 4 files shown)


src/index.template.html
```diff
@@ -249,9 +192,16 @@
...
           }
         }
-      //If in 5 seconds it hasn't render the root of the page thne reload the page.
+      //If in 5 seconds it hasn't render the root of the page then reload the page.
       setTimeout(forceRefreshIfNotloaded, 30000);
+
+      window.addEventListener('load', function (event) {
+        if (!('noModule' in HTMLScriptElement.prototype)) {
+          document.getElementById('error').innerHTML =
+            "<b>⚠️ Error!</b> The version of this browser doesn't support some essential features used in InterPro.<br/><br/>Please upgrade your browser.";
```
matthiasblum, line 201:
```suggestion
            "Sorry, your web browser does not support JavaScript modules, which are required for InterPro. Please upgrade your browser to a newer version, or switch to a different browser that supports JavaScript modules.";
```

src/index.template.html
```diff
@@ -249,9 +192,16 @@
...
+          console.error('No support for es-modules');
```
matthiasblum, line 202: Is it necessary for end-users?

src/pages/Home/index.js
```diff
@@ -225,6 +247,17 @@ class Home extends PureComponent {
             settingsName="showSettingsToast"
           />
         ) : null}
+        {'noModule' in HTMLScriptElement.prototype && supportsES6() ? null : (
+          <ul className={f('toast-display')}>
+            <Toast
+              title="⚠️ Unsuported Browser"
+              body="We have detected some missing features in your browser. You might experience some issues in our website. Please update your browser."
```
matthiasblum, line 254:
```suggestion
              body="We have detected that your web browser does not support JavaScript modules, which are required for the InterPro website to function properly. As a result, certain features may not work as expected. We recommend that you upgrade your browser to the latest version or switch to a supported browser such as Firefox, Chrome, or Edge."
```

src/index.template.html
matthiasblum: Well, maybe the error message is not required since there is a toast...

gustavo-salazar: It's actually needed for the extreme case where none of the website JS gets loaded.

src/index.template.html
```diff
@@ -249,9 +192,16 @@
             window.location.reload();
           }
         }
-      //If in 5 seconds it hasn't render the root of the page thne reload the page.
+      //If in 5 seconds it hasn't render the root of the page then reload the page.
       setTimeout(forceRefreshIfNotloaded, 30000);
```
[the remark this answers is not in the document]
gustavo-salazar: `30 === 5`😑

src/index.template.html
gustavo-salazar: no necessary, but thought of leaving it there, for debugging RT tickets.

PR #474, opened 2023-04-17
tigrfams => ncbifam

gustavo-salazar:
Most of the logic to support a new member DB is already in place as mostly everything is dynamically generated from the API responses. A few things were still necessary:

* Adding a color definition for `ncbifam` in the config and in some css files.
* Adding images for the logo of `ncbifam`. In this case using the NIH>NLM logo which is the parent institution of NCBI.
* Adding `NCBIfam` as one of the application for interproscan advance search.
* Adding a Redirect component in `<Pages>` to detect if `customLocation.entry.db ===  "entry"` and then redirect anything to a URL with `ncbifam` instead.
* Adding a regular expresion for `NCBIfam` accessions.
* Defining the URL to direct external links to NCBIfam


matthiasblum (approved):
Looks good to me. Thanks!

PR #477, opened 2023-05-17
Minor fixes of latest deployed version

gustavo-salazar:
A couple of minor fixes:
* The set viewer was breaking the page when the headers changes to sticky
* Some logic to manage the cases where an entry doesn't have descriptions, but has citations. Currently, it looks like 2 sections of references, with this PR it looks like the ones from the member DBs are additional.

**Note 1:** There are some dependencies updates so remember to `npm install`
**Note 2:** @typhainepl please check if the conservation checkbox works now. I think it does, but the only thing I did was to update the nightingale component. So this might need more work.


src/components/Description/DescriptionFromIntegrated/index.tsx
```diff
@@ -33,7 +66,12 @@ interface IntegratedProps
 const DescriptionFromIntegrated = ({
   integrated,
   data: { loading, payload },
+  setIntegratedCitations = (_: string[]) => null,
 }: IntegratedProps) => {
+  useEffect(() => {
+    setIntegratedCitations(Object.keys(payload?.metadata?.literature || {}));
+    console.log(Object.keys(payload?.metadata?.literature || {}));
```
typhainepl, line 73: This line can be removed for production

typhainepl (commented):
Looking good to me. Just one log print to remove.

As you are there, would it be possible to remove the `^` symbol displayed after the references number? e.g. change `1.^` to `1.`

typhainepl:
Just noticed that the changes broke the InterPro entry references e.g. http://localhost:8080/interpro/entry/InterPro/IPR001279/

PR #489, opened 2023-06-23
Domains on protein TypeScript migration

gustavo-salazar:
#### TypeScript migration of the following components:
- `Genome3D`
- `DomainsOnProtein` - Split into multiple files:
  - `DomainsOnProteinLoaded`
  - `ConservationProvider`
  - `mergeExtrafeatures`
  - `mergeResidues`

A new High Order Component `ExternalSourcesHOC` was created to group all the external sources to add to the protein viewer. At the point of this PR only Genome3D is set up as an external source. RepeatsDB and DisProt will be added in a separate PR.

#### Multiple typing improvements.
- Some ProteinViewer types moved into `custom.d.ts` to make them available all around
- The `loadData` function nos has another generic on its type definition to be able to include Namespace, and this way `loadData` can be nested and its wrapped component will have the types well defined for the prop `data${Namespace}`
- `data` in `LoadDataPropsBase` can now be `undefined` when the data hasn't been loaded yet. This needed minor changes in multiple files

#### Other changes:
- `ProteinViewer/Options` didn't need a `loadData`
- The calculate conservation is now disabled until HmmerWeb is working again


src/components/IPScan/Summary/index.js
```diff
@@ -23,7 +23,8 @@ import GoTerms from 'components/GoTerms';
 import Accession from 'components/Accession';
 import ProteinEntryHierarchy from 'components/Protein/ProteinEntryHierarchy';
 import Length from 'components/Protein/Length';
-import { DomainOnProteinWithoutMergedData } from 'components/Related/DomainsOnProtein';
+// $FlowFixMe
```
typhainepl, line 26: What's the purpose of this comment?

typhainepl (commented):
Looking good, however I can't test the Genome3D annotations locally, what configuration should I set to get the data?
I spotted a glitch when changing the colouring selection for [A0A140JWM8](http://localhost:8080/interpro/protein/reviewed/A0A140JWM8/#table) and [O00418](http://localhost:8080/interpro/protein/reviewed/O00418/) but it doesn't happen for [A0A0G2JDV3](http://localhost:8080/interpro/protein/reviewed/A0A0G2JDV3/#table).

src/components/IPScan/Summary/index.js
gustavo-salazar: The short answer is that is a way to ignore Flow errors.

We are migrating away from FlowJS and implementing all the type validation with TypeScript. But because we are doing it gradually there are files still using Flow that are now importing TS files. Which causes an error in flow, and that's why you see me putting so many of this comments in the migrations PRs

src/components/Related/DomainsOnProtein/index.tsx
```diff
@@ -0,0 +1,282 @@
...
+        getUrl: getExtraURL('extra_features'),
+        propNamespace: 'Features',
+      } as Params)(
+        loadData<ResiduesPayload, 'Residues'>({
+          getUrl: getExtraURL('residues'),
+          propNamespace: 'Residues',
+        } as Params)(
+          loadData(getRelatedEntriesURL as Params)(DomainOnProteinWithoutData)
+        )
+        //     ),
```
matthiasblum, line 278:
```suggestion
```

matthiasblum (approved):
Looks good to me.

PR #500, opened 2023-07-11
Antifam available as a memberDB

### LINKED ISSUE #498: Remove "Add your annotation" from signatures pages except Pfam
ATM, the "Add your annotation" button is found in all signatures pages, however we can't update those, except Pfam. My suggestion is to remove it except for Pfam signatures pages.

gustavo-salazar:
Changes to use AntiFam similarly to a member database:
* Protein viewer as an unintegrated match
* Entry subpages for a protein with an AntiFam match
* As an option in the `Browse By Member DB`
* AntiFam entries have their own page as other member DB signatures
* Assigning color  `#b89494`  to AntiFam

The only place that behaves differently than a member DB is the home page where it's excluded from the list of member DBs.

**Note:** requires an API that has been configured to include AntiFam

The last commit also FIX #498


matthiasblum (approved):
Looks good to me.

The two issues I found are actually something I need to deal with when building the DW:

- exclude AntiFam from the release notes
- add the HMMs so they can be viewed/downloaded

typhainepl:
What's the API link to test?

matthiasblum:
> What's the API link to test?

Until the API changes are merged  (see ProteinsWebTeam/interpro7-api#111) then deployed to the dev API VM (`wp-np3-af`) you need to have a local API (checked out on `antifamDB`). I'll see with @gustavo-salazar on Monday if we can merge/deploy already.

> Do you have a protein example?

* A0A009H0X5
* A0A645G3X3
* F2A9W7
* K7BFE2
* W9B4C4

gustavo-salazar:
@typhainepl just deployed the dev API so you can use that one to test.
For instance: http://wp-np3-af.ebi.ac.uk/interpro/api/entry/antifam/

typhainepl (commented):
The Browse works fine.
However, in the protein page, the Entries tab is shown and says there is one entry, when there isn't any in reality.
In the AntiFam signature page we shouldn't display the external link. 
I am not sure about the colour scheme of the AntiFam accession, the black text on grey background isn't very visible, can you use `#b89494` instead?

gustavo-salazar:
> However, in the protein page, the Entries tab is shown and says there is one entry, when there isn't any in reality.

This was a bug for any protein with matches but none integrated. I change the link for those cases to `/all` instead of /interpro`.

>In the AntiFam signature page we shouldn't display the external link.

OK.

> I am not sure about the colour scheme of the AntiFam accession, the black text on grey background isn't very visible, can you use #b89494 instead?

Isn't that the color I'm using? I'll make it a bit lighter (`#d8a4a4`) in any case

typhainepl:
> > I am not sure about the colour scheme of the AntiFam accession, the black text on grey background isn't very visible, can you use #b89494 instead?
> 
> Isn't that the color I'm using? I'll make it a bit lighter (`#d8a4a4`) in any case

I mean the colour used for the accession text background, I've got the following ATM:

gustavo-salazar:
Oh I see. there is a config file that is not in github: `/config.yml`
add `  antifam: '#d8a4a4'` at the end where the colors of the memberDBs are. and restart the server.

typhainepl:
> Oh I see. there is a config file that is not in github: `/config.yml` add ` antifam: '#d8a4a4'` at the end where the colors of the memberDBs are. and restart the server.

Got it! It's working fine now

PR #501, opened 2023-07-11
Tidying up other features

gustavo-salazar:
Tidying up the labels for `Other features` in the protein viewer.

* Tracks are sorted by source database, so all the features from the same resource are now together.
* The consensus disorder annotations are on top of MobiDB lite ones.
* All annotations now have the source database as part of the label, making it clear where annotations are coming from.
* Tries to use `seq_feature` for the label with `accession` as fallback.


src/components/ProteinViewer/LabelsInTrack/ExceptionalLabels/index.tsx
matthiasblum, line 1: Ideally, the `seq_feature` wouldn't be used for MobiDB-lite tracks only, but also all tracks with an `ExceptionalLabel`.

Right now, most extra features do not have a value for `seq_feature`, but in the future they might.

matthiasblum (changes requested):
Thanks. Couple of things to change, if possible:

- check for `seq_feature` for all extra features, not just MobiDB-lite
- make labels consistent between web and printing modes (e.g. prefix is _Pfam-N:_ in web mode, but _N:_ in printing mode)

matthiasblum (approved):
lgtm

PR #505, opened 2023-07-21
Single domain track

gustavo-salazar:
Adds a new category in the protein viewer: **Representative Domains**

Uses a similar algorithm to the one provided by @typhainepl in [the demo](https://github.com/ProteinsWebTeam/pronto/blob/c296c6a987e6d0d999b9dbfd90266f955d6efd17/pronto/api/single_dom.py).

Doesn't have a label on the right, any suggestions for it are welcome, but I thought that given the title of the category, a label wasn't necessary.

### DIFF
--- a/src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
+++ b/src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
@@ -52,7 +52,16 @@ const RepresentativeDomainsTrack = ({
     setData(
       entries.map((entry) => ({
         ...entry,
-        color:  getTrackColor(colorDomainsBy===EntryColorMode.DOMAIN_RELATIONSHIP?({parent:{accession:entry.integrated}}):entry, colorDomainsBy),
+        color: getTrackColor(
+          colorDomainsBy === EntryColorMode.DOMAIN_RELATIONSHIP
+            ? {
+                parent: entry.integrated
+                  ? { accession: entry.integrated }
+                  : null,
+              }
+            : entry,
+          colorDomainsBy
+        ),
       }))
     );
   }, [entries, colorDomainsBy]);
--- a/src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
+++ b/src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
 
 import NightingaleInterProTrack from 'components/Nightingale/InterProTrack';
+import LabelsInTrack from 'components/ProteinViewer/LabelsInTrack';
 import DomainPopup from 'components/ProteinViewer/Popup/RepresentativeDomain';
 import { getTrackColor, EntryColorMode } from 'utils/entry-color';
 
@@ -24,6 +25,7 @@ type Props = {
   colorDomainsBy?: keyof typeof EntryColorMode;
   openTooltip: (element: HTMLElement | undefined, content: ReactNode) => void;
   closeTooltip: () => void;
+  isPrinting: boolean;  
 };
 
 const RepresentativeDomainsTrack = ({
@@ -34,6 +36,7 @@ const RepresentativeDomainsTrack = ({
   colorDomainsBy,
   openTooltip,
   closeTooltip,
+  isPrinting,  
 }: Props) => {
   const [data, setData] = useState(entries);
   const containerRef = useRef<HTMLDivElement | null>(null);
@@ -86,6 +89,14 @@ const RepresentativeDomainsTrack = ({
           use-ctrl-to-zoom
         />
       </div>
+      {entries.length === 1 ? (
+        <LabelsInTrack
+          entry={entries[0]}
+          hideCategory={hideCategory}
+          expandedTrack={true}
+          isPrinting={isPrinting}
+        />
+      ) : null}
     </>
   );
 };
--- a/src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
+++ b/src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
@@ -1,0 +1,68 @@
+import React, { ReactNode, useRef, useEffect } from 'react';
+
+import NightingaleInterProTrack from 'components/Nightingale/InterProTrack';
+import DomainPopup from 'components/ProteinViewer/Popup/RepresentativeDomain';
+
+import { ExtendedFeature } from '..';
+
+import cssBinder from 'styles/cssBinder';
+
+import style from '../style.css';
+import grid from '../grid.css';
+
+const css = cssBinder(style, grid);
+
+type Props = {
+  entries: Array<ExtendedFeature>;
+  highlightColor: string;
+  hideCategory: boolean;
+  length: number;
+  openTooltip: (element: HTMLElement | undefined, content: ReactNode) => void;
+  closeTooltip: () => void;
+};
+
+const RepresentativeDomainsTrack = ({
+  entries,
+  highlightColor,
+  hideCategory,
+  length,
+  openTooltip,
+  closeTooltip,
+}: Props) => {
+  const containerRef = useRef<HTMLDivElement | null>(null);
+  useEffect(() => {
+    containerRef?.current?.addEventListener('change', (event) => {
+      const detail = (event as CustomEvent)?.detail;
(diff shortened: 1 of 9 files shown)


src/components/ProteinViewer/utils.ts
```diff
@@ -1,7 +1,60 @@
...
+          length: end - start,
+          keep: true,
+        });
+      }
+    }
+  }
+  // const sortedDomains = flatDomains.sort((a, b) => b.length - a.length);
+  for (const dom1 of flatDomains) {
+    for (const dom2 of flatDomains) {
+      if (dom1 === dom2 || !dom1.keep || !dom2.keep) continue;
```
matthiasblum, line 41: Let N be the number of domains in `flatDomains`.

This code will perform at most $(N^2 - N)$ comparisons, but $(N^2 - N) \over 2$ would be enough.

```ts
flatDomains.forEach((dom1, i) => {
  flatDomains.slice(i+1).forEach((dom2) => {
    ...
  });
});
```

matthiasblum (commented):
Looks good to me. I made a few comments, but nothing major.

src/components/ProteinViewer/utils.ts
```diff
@@ -1,7 +1,60 @@
...
+          start,
+          end,
+          color: getTrackColor({ source_database }, EntryColorMode.MEMBER_DB),
+          length: end - start,
+          keep: true,
+        });
+      }
+    }
+  }
+  // const sortedDomains = flatDomains.sort((a, b) => b.length - a.length);
```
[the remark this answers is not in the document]
gustavo-salazar: Mhh, I actually commented it because I was not completely sure the sorting was necessary. So I was testing without it, and it worked all right for all the examples Typhaine provided. But I'm still unsure if there are edge cases that I haven't tested. Maybe I should just leave the sorting, just to be sure.

typhainepl:
Looking nice! 
However, the viewer for the following protein: P02144, P00519, O00206, D5KQK4, P21865 isn't loading for me following the latest changes.

typhainepl:
The problem mentioned above persists, looking at the console, it seems to be related to genome3D data trying to be fetched through https://www.genome3d.eu/api where it should be https://www.genome3d.net/api

matthiasblum:
> The problem mentioned above persists, looking at the console, it seems to be related to genome3D data trying to be fetched through https://www.genome3d.eu/api where it should be https://www.genome3d.net/api

[genome3d.net](https://www.genome3d.net) is down as well (reported to @sillitoe), so that's not the cause of the issue you're experiencing. I haven't tried all proteins, but those I've tried work on the test VM and locally. 

The test server queries [genome3d.net](www.genome3d.net), so if you see [genome3d.eu](www.genome3d.eu) in the network tab of the DevTools console, I guess you're running a local API/client, and maybe there is something wrong in your configuration.

gustavo-salazar:
OK I checked the `future` deployment and it's using `www.genome3d.net` however this config use to be editable from the client from the [settings page](http://wp-np3-aa.ebi.ac.uk/future/settings/). 

However, that's not an option anymore, I removed it a while ago, so you might need to remove the saved settings from the LocalStorage of your browser: In chrome developer tools, Storage, Local Storage.


In any case the failure on genome3d shouldn't affect the creation of the single domain track, are you still having issues with those proteins?

gustavo-salazar:
OK got it! Good catch!
I just push a fix for it

src/components/ProteinViewer/index.tsx
```diff
@@ -264,19 +269,30 @@ export const ProteinViewer = ({
...
-                      />
+                      {type === 'representative domains' ? (
+                        <RepresentativeDomainsTrack
+                          hideCategory={hideCategory[type]}
+                          highlightColor={highlightColor}
+                          entries={entries}
+                          length={protein.sequence.length}
+                          openTooltip={openTooltip}
+                          closeTooltip={closeTooltip}
+                        />
```
matthiasblum, line 280:
```suggestion
                          isPrinting={isPrinting}
                        />
```

src/components/ProteinViewer/utils.ts
```diff
@@ -1,7 +1,61 @@
...
+  'hamap',
+  'sfld',
+  'cdd',
+  'profile',
+];
+
+const selectRepresentativeDomains = (domains: Record<string, unknown>[]) => {
+  const flatDomains = [];
+  for (const domain of domains) {
+    const { accession, short_name, source_database, integrated } = domain;
```
matthiasblum, line 20:
```suggestion
    const { accession, short_name, name, source_database, integrated } = domain;
```

src/components/ProteinViewer/utils.ts
```diff
@@ -1,7 +1,61 @@
...
+    for (const location of domain.entry_protein_locations as Array<ProtVistaLocation>) {
+      for (const fragment of location.fragments) {
+        const { start, end } = fragment;
+        flatDomains.push({
+          accession,
+          short_name,
```
matthiasblum, line 26:
```suggestion
          short_name,
          name,
```

src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { ReactNode, useRef, useEffect, useState } from 'react';
+
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+
+import NightingaleInterProTrack from 'components/Nightingale/InterProTrack';
+import DomainPopup from 'components/ProteinViewer/Popup/RepresentativeDomain';
```
matthiasblum, line 7:
```suggestion
import LabelsInTrack from 'components/ProteinViewer/LabelsInTrack';
import DomainPopup from 'components/ProteinViewer/Popup/RepresentativeDomain';
```

src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+const css = cssBinder(style, grid);
+
+type Props = {
+  entries: Array<ExtendedFeature>;
+  highlightColor: string;
+  hideCategory: boolean;
+  length: number;
+  colorDomainsBy?: keyof typeof EntryColorMode;
+  openTooltip: (element: HTMLElement | undefined, content: ReactNode) => void;
+  closeTooltip: () => void;
```
matthiasblum, line 26:
```suggestion
  closeTooltip: () => void;
  isPrinting: boolean;  
```

src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+};
+
+const RepresentativeDomainsTrack = ({
+  entries,
+  highlightColor,
+  hideCategory,
+  length,
+  colorDomainsBy,
+  openTooltip,
+  closeTooltip,
```
matthiasblum, line 36:
```suggestion
  closeTooltip,
  isPrinting,  
```

src/components/ProteinViewer/RepresentativeDomainsTrack/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+          margin-color="#fafafa"
+          id="Domains"
+          shape="roundRectangle"
+          highlight-event="onmouseover"
+          highlight-color={highlightColor}
+          show-label
+          label=".feature.short_name"
+          use-ctrl-to-zoom
+        />
+      </div>
```
matthiasblum, line 88:
```suggestion
      </div>
      {entries.length === 1 ? (
        <LabelsInTrack
          entry={entries[0]}
          hideCategory={hideCategory}
          expandedTrack={true}
          isPrinting={isPrinting}
        />
      ) : null}
```

src/components/ProteinViewer/utils.ts
```diff
@@ -1,7 +1,61 @@
...
+          source_database,
+          integrated,
+          start,
+          end,
+          color: getTrackColor({ source_database }, EntryColorMode.MEMBER_DB),
+          length: end - start,
```
matthiasblum, line 32: I didn't see `length` used for anything else than comparing domains, but if it is, it should be:

```suggestion
          length: end - start + 1,
```

matthiasblum (approved):
A few more suggestions. Mainly to display the label next to the representative track (if there is only one representative domain)

PR #509, opened 2023-08-11
Structure sequence

gustavo-salazar:
This PR updates the protein viewers on the structure pages to use the sequence that the PDB chains are defined with. 

#### Changes

- Structure summary now uses the chain sequence, and doesn't depend on existing links with proteins or entries
- Structure subpage on an entry page - now uses the structure sequence, and displays a match track for each chain.
- Structure subpage on a protein page - now uses the structure sequence, and displays a match track for each chain.
- Protein subpage on a structure page - now uses the structure sequence, and displays a match track for each chain.
- Entry subpage on a structure page - now uses the structure sequence, and displays a match track for each chain.

#### Typescript + VF migrations
- `components/Matches`
- `components/Matches/EntriesOnProtein`
- `components/Matches/EntriesOnStructure`
- `components/Matches/StructureOnProtein`
- `components/Related`

#### Other changes
- Removal of the `tippy` library and using `floating-ui` instead. Still missing removing `popper` for a few specific cases.


matthiasblum (changes requested):
Overall, it's good, but a few things can be improved.

**Jump to button**

https://github.com/ProteinsWebTeam/interpro7-client/blob/c63f55a28077f4d3f1b71705e27276b5b2801334/src/components/Related/DomainEntriesOnStructure/GoToProtVistaMenu/index.tsx#L25

Options in the dropdown have empty parentheses if no protein is associated with the chain. Also, jumping to the chain doesn't seem to work.

Example: `/interpro/structure/PDB/7zqa/`

**Relative time**

https://github.com/ProteinsWebTeam/interpro7-client/blob/c63f55a28077f4d3f1b71705e27276b5b2801334/src/components/Structure/Summary/index.tsx#L122-L128

Replace relative time by absolute time. It makes more sense to show `31 Oct 1993` instead of `29 years ago`.

**Performance**

Showing all chains is an overkill, especially if there are many chains (e.g. `7y4l`*). We could show one, and use the drop down to switch chain.

** Redundancy**

If more than one chain is mapped to a protein, show the domains once for all chains, not once per chain (we'd be showing the same domains).

Example: `/interpro/structure/PDB/4nxu/`

**Tooltip**

(Not in the scope of this PR, but I only noticed the issue when reviewing it)

The tooltip that appears when opening the "Export" button is _behind_ the dropdown. Also, the change of the tooltip changes when hovering a link inside.

Example: `/interpro/structure/PDB/4nxu/protein/UniProt/#table`

typhainepl:
I agree with @matthiasblum suggested changes.
Additionally, the secondary structure info isn't displayed for all the chains (this isn't a new issue, it is also present in the public website), e.g. 6m58. I wonder if this is because we don't have the info, or there is a bug somewhere.

Not on the scope of this PR, but problematic when working on a laptop screen, the buttons from the export are partly outside the screen.
image

PR #510, opened 2023-09-11
Structure sequence fix

matthiasblum:
**"Jump To" button**

Hidden. I think there a downstream event listener that breaks the "Jump to" functionality by jumping back to the structure viewer, but I don't have the time to investigate more.

**Relative time**

Showing absolute time instead of relative time (e.g. `31 October 1993` instead of `29 years ago`).

**Performance**

Not done.
I tried to transform the "Jump To" button into a selector so we only show one protein viewer (for one chain) at a time. It works at first, but then there is an unexpected behavior when "cycling back" to the first selected chain (e.g. `A` selected -> switch to `B` -> switch back to `A`).

**Redundancy**

Not done. We need to be able to group chains together based on their sequence. Although this could be done directly in the client (we have the sequence for each chain), it's probably better to do that in the data warehouse.

**Missing secondary structures**

Fixed. Example : 

- missing on the live website: https://www.ebi.ac.uk/interpro/structure/PDB/6m58/
- present on local build at `/interpro/structure/PDB/6m58/`

**Export button**

* Tooltip behind dropdown panel: fixed
* Dropdown overflow viewport on small screens: fixed; the dropdown content is aligned with the right of the "Export" button

Screenshot_2023-09-11_11-49-23

**Other changes**

* Picture-In-Picture for structure viewer disabled by default
* Domain viewer labels: accession and short name used by default (currently, only the accession is used)

### DIFF
--- a/src/components/EntryMenu/EntryMenuLink/style.css
+++ b/src/components/EntryMenu/EntryMenuLink/style.css
@@ -1,4 +1,4 @@
-@import '../../../styles/timing.css';
+@import "../../../styles/timing.css";
 
 :root {
   /*GAL*/
@@ -12,7 +12,7 @@ body {
     &:focus,
     &:hover,
     &:visited {
-      text-decoration: none;
+      text-decoration: none !important;
       font-size: 0.9rem; /* same size as submenu */
       padding: var(--menuitem-padding, 0.4rem);
       color: black;
--- a/src/components/File/TooltipContent/style.css
+++ b/src/components/File/TooltipContent/style.css
@@ -14,5 +14,7 @@
   &:focus {
     color: var(--colors-almost-white);
     box-shadow: none;
+    text-decoration: none !important;
+    border: 4px solid transparent !important;
   }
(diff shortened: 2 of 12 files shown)


typhainepl (commented):
It's looking good to me overall. My only concern is that when the short name is too long (e.g. InterPro entries), the accession is truncated on the side, do you think you could slightly reduce the font (9px seems to work in most cases, but it might be too small) or add a slider at the bottom to display the full text?

matthiasblum:
> add a slider at the bottom to display the full text?

There is already a scrollbar (hidden by default, you need to hover the label for it to appear).

typhainepl:
I can see the scrollbar in Firefox, but it never shows on Chrome.

matthiasblum:
_Sigh... Chrome..._

chrome

The scrollbar should be visible now.

typhainepl:
Yes it is visible, thanks for fixing it. 
You can merge the PR.

PR #511, opened 2023-09-12
PRINTS & SFLD messages

matthiasblum:
Adds a message for PRINTS and SFLD on the "Browse by member DB" with a link to Read the Docs.

Also removes the message about Pfam on the home page.


typhainepl (approved):
Looking good to me.

Issue #518, opened 2023-09-22
Set a maximum size to tooltips


matthiasblum:
Most tooltips are sized automatically based on their content, which leads to over-wide tooltips.

Examples: https://www.ebi.ac.uk/interpro/entry/InterPro/IPR000001/

- Overlapping homologous superfamilies
- Add your annotations
- EC (cross references section)

typhainepl:
Another example: https://www.ebi.ac.uk/interpro/entry/pirsf/PIRSF500625/

PR #519, opened 2023-10-02
Removed signatures

matthiasblum:
Show basic information for deleted signatures and refactor the information displayed for deleted InterPro entries:

- update error message on top
- show the type and source of the deleted entry
- show the short name, if any
- for signatures, show the member database
- show past names and short names, if any

## API

You do not need to have a local API: you can use `http://wp-np3-af.ebi.ac.uk/interpro/api/` in `config.yml`.

## Examples

* `/entry/pfam/PF00256/`
* `/entry/ncbifam/TIGR00108/`
* `/entry/cdd/CD00020/`
* `/entry/InterPro/IPR000005/`

### Live website:

image

### After changes:

image

### DIFF
--- a/src/components/Entry/RemovedEntrySummary/index.js
+++ b/src/components/Entry/RemovedEntrySummary/index.js
@@ -17,7 +17,7 @@ const RemovedEntrySummary = ({
   type,
   name,
   short_name,
-  date,
+  deletion_date,
   history,
   dbInfo,
 }) => {
@@ -39,6 +39,10 @@ const RemovedEntrySummary = ({
         : [],
     is_removed: true,
   };
+  const date = new Date(deletion_date).toLocaleDateString('en-GB', {
+    month: 'long',
+    year: 'numeric',
+  });
   const message = `${metadata.accession} has been retired in ${date}.`;
   return (
     <div className={f('row')}>
@@ -62,7 +66,7 @@ RemovedEntrySummary.propTypes = {
   type: T.string,
   name: T.string,
   short_name: T.string,
-  date: T.string,
+  deletion_date: T.string,
   history: T.shape({
     names: T.arrayOf(T.string),
     short_names: T.arrayOf(T.string),
--- a/src/components/Entry/RemovedEntrySummary/index.js
+++ b/src/components/Entry/RemovedEntrySummary/index.js
@@ -36,7 +36,7 @@ const RemovedEntrySummary = ({
     description:
       formerNames.length !== 0
         ? ['Former names:', `<ul>${formerNames.map(listWrap).join('')}</ul>`]
-        : null,
+        : [],
   };
   const message = `${metadata.accession} has been retired in ${date}.`;
   return (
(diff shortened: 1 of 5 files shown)


src/components/Entry/RemovedEntrySummary/index.js
```diff
@@ -11,28 +11,42 @@ import ebiGlobalStyles from 'ebi-framework/css/ebi-global.css';
...
+    },
+    source_database: source_database,
+    type: type,
     member_databases: history?.signatures,
-    description: [
-      `<b>Removed</b>: ${date}`,
-      '<b>Used names</b>:',
-      ...(history?.names || []).map((n) => ` * ${n}`),
-    ],
+    description:
```
typhainepl, line 36: can you remove the `description` title?

src/components/Entry/Summary/SidePanel/index.tsx
```diff
@@ -96,80 +96,87 @@ const SidePanel = ({
...
-          >
-            <DropDownButton
-              label="Add your annotation"
-              icon="&#xf303;"
-              extraClasses={css('annotation')}
+      ) &&
+        !metadata.is_removed && (
+          <div>
+            <Tooltip
+              title={
```
typhainepl, line 103: The tooltip is too long, can you make it on multiple lines?

typhainepl (commented):
It is looking good to me, just a few minor changes.

Regarding the API, I think we should change "date" by "deletion date" to make it clearer that the entry has been deleted.

src/components/Entry/RemovedEntrySummary/index.js
matthiasblum: Done.

src/components/Entry/Summary/SidePanel/index.tsx
matthiasblum: That's a site-wide issue, reported in #518.
I may be able to fix it later this month, otherwise @gustavo-salazar will take care of it.

matthiasblum:
> Regarding the API, I think we should change "date" by "deletion date" to make it clearer that the entry has been deleted.

Done.

typhainepl (commented):
Looking good!
I discussed with Sara and we think it will be better to say `deleted` than `retired` in the statement.

matthiasblum:
> I discussed with Sara and we think it will be better to say `deleted` than `retired` in the statement.

_Deleted_ implies that the resource is no longer available. _Retired_ means it's no longer used or updated but may still exist. I think the later applies more to signatures we do not have anymore. For instance `CD00020` is not used in InterPro and InterProScan anymore, but is still available on the CDD website at https://www.ncbi.nlm.nih.gov/Structure/cdd/cddsrv.cgi?uid=CD00020

typhainepl:
> > I discussed with Sara and we think it will be better to say `deleted` than `retired` in the statement.
> 
> _Deleted_ implies that the resource is no longer available. _Retired_ means it's no longer used or updated but may still exist. I think the later applies more to signatures we do not have anymore. For instance `CD00020` is not used in InterPro and InterProScan anymore, but is still available on the CDD website at https://www.ncbi.nlm.nih.gov/Structure/cdd/cddsrv.cgi?uid=CD00020

How about we say "has been obsoleted"?

typhainepl (approved):

PR #522, opened 2023-10-05
Pfam helpdesk

matthiasblum:
When submitting an annotation from a Pfam entry page, the email is sent to the Pfam helpdesk instead of the InterPro helpdesk.

## API

Local copy on the `dev` branch, or `http://wp-np3-af.ebi.ac.uk/interpro/api/`.


typhainepl (approved):
Looking good and good job on fixing the tooltip issue.

PR #523, opened 2023-10-05
Pfam set/clan terminology

matthiasblum:
Display _clan_ instead of _set_ on Pfam entry and set pages.


src/components/Title/TitleTag/index.tsx
```diff
@@ -27,7 +27,12 @@ type Props = {
 
 const TitleTag = ({ db, mainType, dbLabel }: Props) => {
   const isEntry = mainType === 'entry';
-  const isInterPro = db && db.toLowerCase() === 'interpro';
+  let isInterPro = false;
```
typhainepl, line 30: I'm a bit puzzle by the name of this variable, why `isInterPro`?

typhainepl (commented):
Looking good to me.
However, I think we should also change the label in the `Browse` from `By Set` to `By Set/Clan`.

laisecf (approved):

src/components/Title/TitleTag/index.tsx
matthiasblum: The `TitleTag` component is the label/tag that is at the top of most pages, with the source and type of data being browsed.

_UniProtKB/TrEMBL protein_

image

_PIRSF entry_

image

_InterPro entry_

image

The `isEntry` and `isInterPro` variables are used to:

* select the colour of the tag (`#1878ba` or `#4e7287`)
* generate the link to the correct section on the entry page on docs whether the entry is an InterPro entry or a member database signature.

typhainepl: Thanks for clarifying. I was a bit confused as when I initially read the code, I thought you were checking whether the db was `interpro` (for all the entries and signatures) and thinking in my head that Pfam is also part of InterPro.
After looking at it again, I got what you are trying to do.

typhainepl (approved):

typhainepl (approved):

typhainepl (approved):

typhainepl:
Ready to merge in my opinion

typhainepl (approved):

matthiasblum:
> I think we should also change the label in the `Browse` from `By Set` to `By Set/Clan`.

## Updating the label in the _Browse_ dropdown

Also, the _By Clan/Set_ link redirects to Pfam clans, instead of all clans/sets, which is consistent with the behaviour of the _By Member DB_ link that redirects to the list of Pfam signatures.

image

## Updating the label in the navigation path

Only when browsing Pfam clans.

image

PR #525, opened 2023-10-31
Use pre-selected representative domains

matthiasblum:
Representative domains are now selected during the data warehouse build process, and are available in the REST API. 

## API

Local copy on the `dev` branch, or `http://wp-np3-af.ebi.ac.uk/interpro/api/`.


typhainepl (approved):
Looking good to me :)

laisecf (approved):

PR #527, opened 2023-11-06
Fix endpoint page loading

matthiasblum:
`endpoint-page.js` implements the logic to display the overview summary of primary endpoints, show the basic information if an InterPro entry or member database signature has been removed, or show error messages (for instance if there is no data returned by the API).

I previously, re-ordered some of the conditions, but this caused an issue when open a entry page for a PANTHER subfamily (`204` returned by the API), but since `payload` is null the client kept showing a "loading" state.

This PR fixes the order of conditions to test to show data or redirect.

### DIFF
--- a/src/pages/endpoint-page.js
+++ b/src/pages/endpoint-page.js
@@ -114,22 +114,31 @@ class Summary extends PureComponent {
       },
     } = customLocation;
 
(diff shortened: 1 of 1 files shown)


gustavo-salazar (approved):
looks good to me!

PR #528, opened 2023-11-07
set subpages without counters

gustavo-salazar:
Getting counters from the API is an expensive operation, especially on big lists.

Although this requires more investigation to be able to generalise it, it seems to me that for sets the counters in a page like https://www.ebi.ac.uk/interpro/set/pfam/CL0010/entry/pfam/#table are not necessary. I couldn't find any case where they are currently used.

The code change is minimal, but I'm afraid of any cases I miss, so I'd appreciate a check before merging!


matthiasblum (approved):
Looks good to me.

PR #535, opened 2023-11-14
Remove set alignment and rosettaFold subpages

gustavo-salazar:
We are removing the code and making the following subpages unavailable:
* profile-profile alignments for sets 
* RosseTTaFold models for pfam entries and their integrated InterPro.

Also adding an error handler for such pages, in case a user has a bookmark to go directly to the link. If you are testing locally you can try:
* http://localhost:8000/interpro/entry/pfam/PF16915/rosettafold/
* http://localhost:8000/interpro/set/pfam/CL0001/alignments

TypeScript migrations
* `src/components/EdgeCase`
* `src/components/Progressutton`

Plus some minor eslint fixes.


matthiasblum (changes requested):
Looks nice. I'd suggest to minor improvements:

- prevent the `EdgeCase` callout to take all the viewport width,
- for the RoseTTAFold message, add a link to the AlphaFold page for the same entry.

Both suggestions are implemented in #537, please have a look; if the code is good enough, feel free to merge #537 before merging this PR, otherwise you can implement the changes here.

PR #537, opened 2023-11-15
Dynamic message for deleted pages

matthiasblum:
Enable dynamic messages for deleted pages, essentially to show a link from the RoseTTAFold page to the AlphaFold page.
Also, wrap the callout in a container so it doesn't take the full width of the viewport.


gustavo-salazar (approved):
LGTM!
Good Idea to make the message a react component

PR #540, opened 2023-11-20
removing twitter timelines and just leaving a link

gustavo-salazar:
The twitter widget, hasn't been working properly for a while. 
The issue is with the twitter API an changes they have made. Multiple [forums](https://twittercommunity.com/t/embedded-tweets-not-working/199477/52?page=2) are reporting this sort of failure.

This PR removes the use of the Twitter API, and reduces the component to a link to the [timeline in twitter](https://twitter.com/InterProDB). This however also has an issue, as it seems that latest changes in twitter, require users to be logged in to see the full timeline, otherwise they get only ani pinned tweets. For this reason, I have pinned the latest release tweet from us.


matthiasblum (approved):
Looks good, thanks! Feel free to merge.

Issue #539, opened 2023-11-17
The matches column in the Entries tab of the sequence search is empty


typhainepl:
I have noticed that the `Matches` column in the `Entries` tab found in the sequence search result is now empty. It used to show where the matches were located in the protein.

Issue #534, opened 2023-11-09
The reviewed tick label is displayed below accession instead of next to it in protein lists


typhainepl:
The reviewed tick label is now sometimes displayed below the protein accession instead of next to it in protein lists; it doesn't look nice...
Example:
https://www.ebi.ac.uk/interpro/entry/InterPro/IPR014856/protein/reviewed/#table

Issue #533, opened 2023-11-09
Add link to Pfam documentation in the InterPro website


typhainepl:
It would be great to add a link to the [Pfam documentation](https://pfam-docs.readthedocs.io/en/latest/index.html) in the Help/Documentation section of the website.

Issue #529, opened 2023-11-09
Fix cache not cleared leading to Error message


typhainepl:
A number of users have reported web pages showing an Error message on accessing a wide range of pages. Hard refreshing the page resolves the issue, but we should find a way to have the cache automatically cleared to avoid having this issue in the future, as it is a bad user experience.

PR #541, opened 2023-11-20
Minor fixes

matthiasblum:
This PR addresses a few minor issues/improvements:

- Delete all local caches before reloading, when clicking the _Reload this page_ button #529
- Add a link to the Pfam documentation (visible at `/interpro/help/documentation/`) #533
- Prevent the checked/ticked  icon (:white_check_mark:) to wrap to a new line when browsing reviewed proteins (e.g. `/interpro/protein/reviewed/entry/InterPro/#table`) #534
- Show again individual matches in the _Entries_ subpage for InterProScan search results #539 
- Check for the value/presence of the `LLMFeedback` config option before showing the button to the external Google Form (so it can be easily disabled or no error is raised if the option is missing from the config file)


gustavo-salazar (approved):
Offline discussed changes were implemented, so happy for this to be merged

PR #547, opened 2023-11-30
Ipscan batch download

gustavo-salazar:
With this PR the `Group Actions` dropdown in the list of ipsan results now checks if the job is still in the server. If that's the case it adds Download All options in the available formats: XML, TSV, etc. If the result is not in the server, only the JSON option is presented, which reconstructs a single file from the multiple jobs.

#### Other changes
* Adding `React.memo` to the `Table` component to reduce some of the re-renderings.
* `GroupActions` is now passed as an element instead of a highorder function, which fixes an infinite loop when checking the ipscan result endpoint.
* Removing `row/column` styles in the clan summary

#### Migrate to typescript
* `DropDownButton`
* `Group/Actions`


src/components/IPScan/Actions/Group/DownloadAll/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+};
+interface LoadedProps extends Props, LoadDataProps<string> {}
+
+const DownloadAll = ({ jobs, group, remoteID, data, dataURL }: LoadedProps) => {
+  const handleDownload = async () => {
+    downloadFile(await getAllResults(jobs, group), `${group}.json`);
+  };
+  const thereDataInServers = remoteID && data?.payload === 'FINISHED';
+  return thereDataInServers ? (
+    ['tsv', 'json', 'xml', 'gff', 'sequence'].map((type) => (
```
matthiasblum, line 52: Input before outputs:

```suggestion
    ['sequence', 'tsv', 'json', 'xml', 'gff'].map((type) => (
```

src/components/IPScan/Actions/Group/DownloadAll/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+          }
+        >
+          <Link
+            target="_blank"
+            href={`${dataURL}/${remoteID}/${type}`}
+            download={`InterProScan-${remoteID}.${type}`}
+            className={css('group')}
+          >
+            <span className={css('icon', 'icon-common')} data-icon="&#xf019;" />{' '}
+            Download All [{type.toUpperCase()}]
```
matthiasblum, line 69:
```suggestion
            {type === 'sequence' ? 'FASTA input' : `${type.toUpperCase()} output`}
```

src/components/IPScan/Actions/Group/DownloadAll/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+              servers. This is only available for 7 days after runnin the job.
+            </div>
+          }
+        >
+          <Link
+            target="_blank"
+            href={`${dataURL}/${remoteID}/${type}`}
+            download={`InterProScan-${remoteID}.${type}`}
+            className={css('group')}
+          >
```
matthiasblum, line 67: This seems to work on the TSV output only. For the other formats, it opens a new tab. Can we force the download?

src/components/IPScan/Actions/Group/DownloadAll/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+    ['tsv', 'json', 'xml', 'gff', 'sequence'].map((type) => (
+      <li key={type}>
+        <Tooltip
+          title={
+            <div>
+              This will download the data that was originally loaded to our
+              servers. This is only available for 7 days after runnin the job.
+            </div>
+          }
+        >
```
matthiasblum, line 61: Minor typo, but I don't think we need a tooltip here, tbh.

```suggestion
        <Tooltip
          title={
            <div>
              This will download the data that was originally loaded to our
              servers. This is only available for 7 days after running the job.
            </div>
          }
        >
```

src/components/IPScan/Actions/Group/DownloadAll/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+        }
+      >
+        <button
+          className={css('icon', 'icon-common', 'ico-neutral', 'group')}
+          onClick={handleDownload}
+          data-icon="&#xf019;"
+          aria-label="Download group results"
+        >
+          {' '}
+          Download All [JSON]
```
matthiasblum, line 91:
```suggestion
          JSON output
```

src/components/IPScan/Actions/Group/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+      <DropDownButton label="Group Actions" icon="&#xf03a;">
+        <ul>
+          <li>
+            <Tooltip
+              title={
+                <div>
+                  <b>Delete All Jobs</b>: This will remove the stored data of
+                  all the sequences associated with this job from your browser.
+                  Remember that search results are only retained on our servers
+                  for 7 days
```
matthiasblum, line 117:
```suggestion
                  Delete all the stored data related to the
                  sequences associated with this job from your browser.
                  Please note that search results are stored
                  on our servers for only 7 days.
```

Also, is it relevant to mention the 7 day limit? This action clear the _local_ data...

src/components/IPScan/Actions/Group/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+              }
+            >
+              <button
+                className={css('icon', 'icon-common', 'ico-neutral', 'group')}
+                onClick={handleReRun}
+                data-icon="&#xf019;"
+                aria-label="Re run all sequences"
+              >
+                {' '}
+                Re-run All
```
matthiasblum, line 154:
```suggestion
                Resubmit All
```

src/components/IPScan/Actions/Group/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+                  sequence search for prepopulating all values with the ones
+                  used for this job.
+                </div>
+              }
+            >
+              <button
+                className={css('icon', 'icon-common', 'ico-neutral', 'group')}
+                onClick={handleReRun}
+                data-icon="&#xf019;"
+                aria-label="Re run all sequences"
```
matthiasblum, line 151:
```suggestion
                aria-label="Resubmit all sequences"
```

src/components/IPScan/Actions/Group/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+            group={group}
+            remoteID={getRemoteID(jobs, group)}
+          />
+          <li>
+            <Tooltip
+              title={
+                <div>
+                  <b>Re-run All Sequences</b>: This will take you to the
+                  sequence search for prepopulating all values with the ones
+                  used for this job.
```
matthiasblum, line 143:
```suggestion
                  Open a new sequence search, with all values prepopulated with those previously used for this job.
```

src/components/ProteinViewer/Options/index.tsx
```diff
@@ -102,7 +102,7 @@ const ProteinViewerOptions = ({
             base.appendChild(style);
           } else
             console.warn(
-              "Couldn't setups the style for the nightingale-saver snapshot "
+              "Couldn't setups the style for the nightingale-saver snapshot ",
```
matthiasblum, line 105:
```suggestion
              "Couldn't set up the style for the nightingale-saver snapshot ",
```

src/components/ProteinViewer/Options/index.tsx
```diff
@@ -172,59 +172,66 @@ const ProteinViewerOptions = ({
...
-                </ul>
-              </li>
-              <hr />
-              <LabelBy />
-              <hr />
-              {/* <li>
+
+        <DropDownButton
+          label={
+            <Tooltip title={'More options to customise the protein viewer'}>
```
matthiasblum, line 178:
```suggestion
            <Tooltip title={'Additional customization options'}>
```

src/components/SimpleCommonComponents/DropDownButton/style.css
```diff
@@ -11,7 +11,7 @@
     cursor: pointer;
   }
   & .dropdown-button {
-    background: var(--colors-interpro);
+    background: var(--colors-nearly-white);
```
matthiasblum, line 14: Why not white?

```suggestion
    background: #ffffff;
```

src/components/IPScan/Actions/Group/DownloadAll/index.tsx
gustavo-salazar: The attribute `download` in a `<a>` tag makes it possible to start a download with the given name of the file. however, if I remember correctly it doesn't work in localhost, just when the URL has a domain or something like that.

gustavo-salazar:
> More generally, I wonder if there is a value in storing results locally. This seems to complicate things more it makes them easier. Couldn't we only store job identifiers, and remove them from the local cache after seven days (or if the user manually delete a job)?

I would argue that is a useful feature to be able to keep the jobs you wanted for longer than the 7 days. Moreover, the logic of uploading files shares the strategy of keeping the file in IDB. So if we want to keep the functionality of uploading a ipscan result file, we would need to keep most of the IDB logic in any case.

matthiasblum:
> > More generally, I wonder if there is a value in storing results locally. This seems to complicate things more it makes them easier. Couldn't we only store job identifiers, and remove them from the local cache after seven days (or if the user manually delete a job)?
> 
> I would argue that is a useful feature to be able to keep the jobs you wanted for longer than the 7 days. Moreover, the logic of uploading files shares the strategy of keeping the file in IDB. So if we want to keep the functionality of uploading a ipscan result file, we would need to keep most of the IDB logic in any case.

Fair enough. I forgot about the fact users can upload their results to visualize them.

Happy with the changes. Feel free to merge. :rocket:

matthiasblum (approved):
Happy with the changes. Feel free to merge. :rocket:

PR #548, opened 2023-12-04
Representative domains for interproscan results

gustavo-salazar:
When the interproscan result includes the boolean property `representative` in the `locations` array, an array of the representative is created and displayed as another category.

I think is safe to deploy this feature even if the ipscan job dispatcher is not updated, because if the flag is not there, then the logic keeps working as before.

Unfortunately, this has only been tested with a single file(attached below.), so maybe better to wait until we have versions of ipscan that generates the flag??

[A0A011TAH1.fasta.json](https://github.com/ProteinsWebTeam/interpro7-client/files/13548648/A0A011TAH1.fasta.json)

TypeScript migration: `serializers.ts`

### 👉🏽 UPDATE
The InterProScan instance in `wwwdev` was updated with the representative domains code, so you can point to it in the settings to test this feature.


matthiasblum (approved):
Thanks, looks good to me! Feel free to merge.

typhainepl:
I noticed that when selecting the Colour by "Domain relationship" the RP is shown in grey and the domain colour isn't used. Can you try to fix it before merging?

typhainepl (approved):

gustavo-salazar:
> I noticed that when selecting the Colour by "Domain relationship" the RP is shown in grey and the domain colour isn't used. Can you try to fix it before merging?

I added a fix for this, please check and let me know if everything goes alright

typhainepl:
It's working fine now, thanks for fixing it.

PR #550, opened 2023-12-11
Removing the genome3d subpage from an entry page

gustavo-salazar:
The Genome 3D annotations are outdated and their servers keep going down. Therefore, we decided to remove the subpage to display the reported models related to an InterPro entry.

This PR removes that code and includes the URLs on the list of removed pages to display an adequate error message if someone directly goes to the link.

It also removes the links in help and documentation about genome3d in InterPro.


src/utils/removed-pages/index.tsx
```diff
@@ -43,6 +43,34 @@ export const removedPages: Array<{
...
+  {
+    description: {
+      main: { key: 'entry' },
+      entry: { detail: 'genome3d' },
+    },
+    getMessage: (location) => {
+      return (
+        <>
+          Genome3D models for an InterPro entry have been removed in InterPro
+          98.0. You can use{' '}
```
matthiasblum, line 55:
```suggestion
          Genome3D models have been removed in InterPro 98.0.
          You can use{' '}
```

matthiasblum (approved):
Looks good to me! Made a tiny suggestion regarding an error message.
Feel free to merge.

PR #554, opened 2023-12-18
New description structure

gustavo-salazar:
With the removal of `llm_description`, a change in the format of entry descriptions was necessary. This PR deals with the new provided format.

The new format is described in TS as:
```typescript
type StructuredDescription = {
  text: string;
  llm: boolean;
  checked: boolean;
};
type EntryMetadata ={
  //...
  description: Array<string|StructuredDescription>
}
```
I'm keeping the option of an array of `string` because the same component is used to display the description of proteins, and other entities.

The assumptions mentioned before have been discarded.

The `Description` component checks per paragraph if it has been generated and checked by a curator, then adds an icon at the top of the paragraph indicating it.

The `DescriptionLLM` component is now only the warning message, and the logic for dealing paragraph by paragraph is in `Description`.

The `dev` API (http://wp-np3-bc.ebi.ac.uk/interpro/api/) is using a DB that has descriptions in the structured way. I have manually edited the following entries for testing purposes: 
* [IPR000003](http://wp-np3-bc.ebi.ac.uk/interpro/api/entry/interpro/ipr000003): 2 blocks, both tagged as LLM, one of them as checked. The blocks have internal `<p></p>` making it look like 4 blocks.
* [IPR000011](http://wp-np3-bc.ebi.ac.uk/interpro/api/entry/interpro/ipr000011): 3 blocks. first one `llm` but unchecked, second one from curators, and third one `llm` and `checked`

You can also use the panther entries for testing


src/components/Description/DescriptionLLM/index.tsx
```diff
@@ -1,99 +1,101 @@
...
+          }}
+          className={css(
+            'small',
+            'icon',
+            'icon-common',
+            'icon-exclamation-triangle',
+          )}
+        />{' '}
+        <p>
+          Some paragraphs of this description have been automatically generated
```
matthiasblum, line 43: _Some paragraphs_ implies more than one paragraph. I tried to come up with a phrasing that works in all cases:
 
```suggestion
          The description below includes sections that have been automatically generated
```

src/components/Description/DescriptionLLM/index.tsx
```diff
@@ -1,99 +1,101 @@
...
+            <span
+              className={css('small', 'icon', 'icon-common', 'icon-magic')}
+            />{' '}
+            at the top of each paragraph to identify generated paragraphs. An
+            additional{' '}
+            <span
+              className={css('small', 'icon', 'icon-common', 'icon-check')}
+            />{' '}
+            icon indicates that a curator has approved the generated text.
+          </b>{' '}
```
matthiasblum, line 64: Using text instead of icons (see other suggestion) so we don't need additional text in the callout.

```suggestion
```

src/components/Description/index.tsx
```diff
@@ -21,30 +37,52 @@ class Description extends PureComponent<Props> {
...
+                        className={css('icon', 'icon-common', 'icon-magic')}
+                      />
+                    </Tooltip>
+                  )}{' '}
+                  {checked && (
+                    <Tooltip title="A curator has checked the content of this generated paragraph">
+                      <span
+                        className={css('icon', 'icon-common', 'icon-check')}
+                      />
+                    </Tooltip>
```
matthiasblum, line 68: Showing the source of the description block using badges instead of icons:

```suggestion
                <div className={css('content', { llm, checked })} key={`${i}_${j}`}>
                  {llm ? (
                    <span className={css('vf-badge', 'vf-badge--tertiary')}>
                      AI-generated
                      <span className={css('details')}>{checked ? 'Reviewed' : 'Unreviewed'}</span>
                    </span>
                  ) : (
                    <span className={css('vf-badge', 'vf-badge--tertiary')}>
                      Expert-curated
                    </span>
```

Instead of using a colored background, I'd suggest a colored left border when the description is AI-generated _and_ unreviewed (see changes to `Description/styles.css`).

src/components/Description/style.css
```diff
@@ -23,3 +23,17 @@ a.text-high:focus {
...
+
+div.llm.checked {
+  background-color: rgb(236, 255, 224, 0.3);
+}
+div.llm .icon {
+  color: lightslategray;
+  &:hover {
+    color: inherit;
+  }
+}
```
matthiasblum, line 39:
```suggestion
.description .content {
  border-left: 4px solid transparent;
  padding-left: 1em;
}
.description .content.llm:not(.checked) {
  border-left-color: #f0ad4e;
}
.vf-badge--tertiary {
  background: #e8e8e8;
  border-color: transparent;
  border-radius: 0.3em;
  color: rgba(0,0,0,.6);
  font-size: .85em;
  text-transform: none;
  user-select: none;
}
.vf-badge--tertiary:hover, .vf-badge--tertiary:visited {
  color: rgba(0,0,0,.6);
}
.vf-badge > .details {
  margin-left: 1em;
  opacity: .8;
}
```

src/components/Description/index.tsx
```diff
@@ -8,11 +8,27 @@ import Paragraph from './Paragraph';
...
+const css = cssBinder(styles, fonts);
 
+export const getDescriptionText = (
+  description: string | StructuredDescription,
+) => (typeof description === 'string' ? description : description.text);
+
+export const hasLLMParagraphs = (
+  description: Array<string | StructuredDescription>,
+): boolean => {
+  if (!description?.length) return false;
```
matthiasblum, line 23: Isn't `description` guaranteed to have a `length` property given its type?

```suggestion
  if (description.length === 0) return false;
```

src/components/Description/index.tsx
```diff
@@ -8,11 +8,27 @@ import Paragraph from './Paragraph';
...
+  if (typeof description?.[0] === 'string') return false;
```
matthiasblum, line 24: Since we just checked the length of `description`, can't we assume it has _at least_ one item?

```suggestion
  if (typeof description[0] === 'string') return false;
```

src/components/Description/index.tsx
```diff
@@ -8,11 +8,27 @@ import Paragraph from './Paragraph';
...
+  const llmDescrpitions = description.filter(
+    (desc) => (desc as StructuredDescription).llm,
+  );
+  return !!llmDescrpitions.length;
```
matthiasblum, line 28: Readability?

```suggestion
  return llmDescrpitions.length > 0;
```

 (Well, more a personal preference tbh)

matthiasblum (commented):
Happy with most of the changes. I have some doubts with the design, but this is something that is highly subjective.

Instead of using icons/tooltips to describe the status of AI-generated blocks, maybe we could use badges so:

- users do not need to hover the icon to know what it means
- we do not need to define the meaning of icons in the top callout (which is already big enough)

Also, I'd suggest not to use different background colors. Instead, we could show a "warning" border for blocks that might contain erroneous information (AI-generated and not reviewed by curators).

Example for `IPR000011`:

image

Please have a look at the suggestions to see if it's something you agree with.

gustavo-salazar:
@matthiasblum cool, I like it, and commited the suggestions.

We were actually talking about what would be the best strategy to display this with @typhainepl so thanks for coming up with an alternative.

matthiasblum (approved):
Looks good to me. :+1:

gustavo-salazar:
One question? @matthiasblum @typhainepl 
In this implementation, the `Expert curated` badge appears on entries where all the paragraphs are curated or taken from the member database.
I feel it's a bit redundant to have it all the time, and maybe even not true in the case of signatures as we have no idea how they created their descriptions.

So I suggest, only showing the badges, when at least one of the paragraphs is AI-generated, what do you think?

matthiasblum:
> So I suggest, only showing the badges, when at least one of the paragraphs is AI-generated, what do you think?

Fine with me.

PR #559, opened 2024-01-06
Fix EntryOnStructure popup

### LINKED ISSUE #556: Error message when hovering over protein matches in structures tab of signature page.
An error message appears when hovering over the protein matches feature in the `Structures` tab of a signature page (e.g. https://www.ebi.ac.uk/interpro/entry/panther/PTHR23411/structure/PDB/#table).

matthiasblum:
Fixes #556 (which occurs when a signature does not have a short name) by not passing the `entry.name` object to `EntryPopup`.


gustavo-salazar (approved):
LGTM

PR #560, opened 2024-01-10
Npm update

gustavo-salazar:
The main change in this PR is the correction of multiple minor issues that were stopping the build from using npm>6

Here is the list of changes:

* Removed unnecessary babel plugins and polifills. Features that are now supported on all modern browsers so there is no need to transpile or polifill.
* Removed storybook dependencies, dealing with that separately.
* Removed `flow-coverage-report` as we haven't used it in years.
* `react-diff-viewer` changed to `react-diff-viewer-continued` a fork that supports react 18
* `react-html-parser` changed to `@hedgedoc/html-to-react` to support react 18 as peer dependency.
* Removed multiple dependencies that were not in use anymore. e.g. webpack plugins not required in webpack 5.
* Remove `legacy-module-split-plugin` as we are only distributing 1 built for modern browsers.
* Remove functional tests `/tests/`. We haven't run them in years and they were never stable.
* Removed scripts related to an old attempt to provide a server-side rendered first page. The approach was never put in place and the improvements shown were marginal.
* Tidied the scripts in `package.json` removing unused ones.
* Updating GH actions workflow to run using the node and npm versions installed in the VMs.
* Using the same version of `core-js` on all dependencies via `override` to avoid multiple warning messages.
* Updating `flow-typed` definitions. Sorry for all the clutter this creates. You can ignore anything under `flow-typed/` for the code review.
* Removing flow checks in several files related to `redux` and `reselect` as these dependencies are not providing Flow types in their new versions. Hopefully, we get to migrate this to TS soon.
* All dependencies were updated to their latest version to date, except for:
  * `flow-bin`: We are stuck on version `0.171.0` for a good while. Any updates will generate hundreds of errors as new constraints were added. And instead of fixing this, we are migrating the code base to TS.
  * `favicons`: The latest version requires a node version that is not available on the cluster VMs. I asked and got his [response from the lib creators](https://github.com/lovell/sharp/issues/3939#issuecomment-1884864830)

A manually deployed instance of this code is now running in http://wp-np3-bd/future/


package.json
```diff
@@ -11,48 +11,38 @@
...
-    "postinstall": "husky install; patch-package",
-    "prepublishOnly": "pinst --disable",
-    "postpublish": "pinst --enable"
+    "test": "npm-run-all test:flow test:lint test:lint:ts test:unit",
+    "test:flow": "flow focus-check --merge-timeout=1000 .",
+    "test:lint": "eslint --color --ignore-path .gitignore --ignore-pattern \"flow-typed\" --ignore-pattern \"lighthouserc.js\" --ignore-pattern \"snippets\" \"./**/*.js\"",
+    "test:lint:ts": "eslint --color --config .eslint4ts.js --no-eslintrc ./src/**/*.ts ./src/**/*.tsx",
+    "test:unit": "jest --maxWorkers=2 src",
+    "version": "node --version && npm --version",
+    "postinstall": "husky install; patch-package"
```
matthiasblum, line 45:
```suggestion
    "postinstall": "husky install && patch-package"
```

Or maybe just remove it? Is it used anywhere?

matthiasblum (approved):
LGTM. Built locally and tested a while without finding any issue. Feel free to merge.

package.json
gustavo-salazar: postinstall is one of those magic commands, it runs as its name says after  `npm install`, so its useful to configure stuff in the project at that point, in this case we are using it for 2 things:
* install husky, to have things like lint and prettier run when committing or pushing.
* running `patch-package` which we use to patch the `EBI-Framework 1.4` to edit a bit of its code and avoid some annoying warnings at build.

gustavo-salazar: why is it better `&&` over `;`?

gustavo-salazar: Based on [this explanation](https://superuser.com/questions/619016/what-is-the-difference-between-command-command-and-command-command/619019#619019) I think it is better to use `;` in this case, because even if `husky` fails installing, the `patch-package` will run.

matthiasblum: > postinstall is one of those magic commands, it runs as its name says after `npm install`, so its useful to configure stuff in the project at that point, in this case we are using it for 2 things:
> 
>     * install husky, to have things like lint and prettier run when committing or pushing.
> 
>     * running `patch-package` which we use to patch the `EBI-Framework 1.4` to edit a bit of its code and avoid some annoying warnings at build.

My bad, I didn't realize it was a built-in feature.

> Based on [this explanation](https://superuser.com/questions/619016/what-is-the-difference-between-command-command-and-command-command/619019#619019) I think it is better to use `;` in this case, because even if `husky` fails installing, the `patch-package` will run.

Fair enough.

PR #561, opened 2024-01-10
Gene in protein list and summary

gustavo-salazar:
Displaying the gene in the protein list and summary pages.
For the list requires the version of the API that currently is in`dev`.


matthiasblum (approved):
LGTM. Thanks!

typhainepl (commented):
The `Gene` column in `Browse By Protein` seems to be always empty, is that expected?
For example, `O00167` has the gene `EYA2` displayed in the protein page, but in the `Browse` it doesn't show anything. http://localhost:8080/interpro/protein/reviewed/entry/InterPro/?search=O00167#table

gustavo-salazar:
> The `Gene` column in `Browse By Protein` seems to be always empty, is that expected?

Which API are you pointing at? Currently only `dev` returns the gene in the list.e.g. http://wp-np3-bc.ebi.ac.uk/interpro/api/protein/uniprot

I'll have a look to the Similar Protein page, and its API call to see how feasible it is

typhainepl:
I was using https://wwwdev.ebi.ac.uk/interpro/api/, using the API mentioned it works fine :)

PR #584, opened 2024-02-21
Do not convert external resource name to upper case

matthiasblum:
Stop forcing cross-reference database names to uppercase.


gustavo-salazar (approved):
LGTM

PR #587, opened 2024-02-23
Ts EntryMenu

gustavo-salazar:
#### Migrating to TS:
* `EntryMenu`
* `EntryMenuLink`

#### Extracting to migrate
* `EntryMenuLinkWithoutData`

#### Type changes
The values in an InterProLocation's Endpoint can be null, as that's how we have them in the redux state. For example `customLocation.description.entry.accession` can be `null` in browse pages.
Multiple files needed an update in order to adjust to that change.


matthiasblum (approved):
LGTM

PR #588, opened 2024-02-26
LLM entries

gustavo-salazar:
This PR is a way to represent the LLM tagged entries.

Basically, I'm reusing the same styling as with LLM descriptions:
* The same type of badge is now displayed in the top of the summary of an entry tagged as LLM, including the `reviewed` or `unreviewed` part.
* I'm extending the Callout box, and changing the content depending if an LLM was used to generate the description, the entry details, or both.

To test this PR I deployed the `dev` API using the `dw-changes` branch: http://wp-np3-bc.ebi.ac.uk/interpro/api/

I also manually changed some of the panther entries in the DB to use them as test data:
* PTHR10000: Entry AND description AI generated unreviewed
* PTHR10003: Entry AI generated reviewed - Description imported from interpro but not LLM
* PTHR10005: Entry AI generated unreviewed - Description imported from interpro but not LLM
* PTHR10009: Description imported from interpro, which was in turn LLM-generated

### DIFF
--- a/src/components/Entry/BadgeAI/style.css
+++ b/src/components/Entry/BadgeAI/style.css
@@ -1,0 +1,17 @@
+.vf-badge--tertiary {
+  background: #e8e8e8;
+  border-color: transparent;
+  border-radius: 0.3em;
+  color: rgba(0, 0, 0, 0.6);
+  font-size: 0.85em;
+  text-transform: none;
+  user-select: none;
+}
+.vf-badge--tertiary:hover,
+.vf-badge--tertiary:visited {
+  color: rgba(0, 0, 0, 0.6);
+}
+.vf-badge > .details {
+  margin-left: 1em;
+  opacity: 0.8;
+}
--- a/src/components/Description/DescriptionFromIntegrated/index.tsx
+++ b/src/components/Description/DescriptionFromIntegrated/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { createSelector } from 'reselect';
 import { format } from 'url';
 
@@ -59,6 +59,7 @@ type Props = {
   integrated: string | null;
   setIntegratedCitations?: (citations: string[]) => void;
   headerText?: string;
+  handleLLMParagraphs?: (hasLLM: boolean) => void;
 };
 
 interface IntegratedProps
@@ -69,12 +70,19 @@ const DescriptionFromIntegrated = ({
   integrated,
   data,
   setIntegratedCitations = (_: string[]) => null,
+  handleLLMParagraphs,
   headerText,
 }: IntegratedProps) => {
   const { loading, payload } = data || {};
+  const [hasLLM, setHasLLM] = useState(false);
   useEffect(() => {
     setIntegratedCitations(Object.keys(payload?.metadata?.literature || {}));
   }, [payload?.metadata?.literature]);
+  useEffect(() => {
+    const newHasLLM = hasLLMParagraphs(payload?.metadata.description || []);
+    setHasLLM(newHasLLM);
+    handleLLMParagraphs?.(newHasLLM);
+  }, [payload?.metadata.description]);
   if (!integrated) return null;
   if (loading) return <Loading />;
 
@@ -86,16 +94,12 @@ const DescriptionFromIntegrated = ({
       payload.metadata.literature,
       citations,
     );
-    const hasLLM = hasLLMParagraphs(payload.metadata.description || []);
 
     return (
       <>
         <h4>
           {headerText || 'Description'} <ImportedTag accession={integrated} />
         </h4>
-        {hasLLM ? (
-          <DescriptionLLM accession={payload.metadata.accession} />
-        ) : null}
         <Description
           textBlocks={payload.metadata.description}
           literature={included}
--- a/src/components/Description/DescriptionLLM/__snapshots__/test.js.snap
+++ b/src/components/Description/DescriptionLLM/__snapshots__/test.js.snap
@@ -5,7 +5,112 @@ exports[`<DescriptionLLM /> should render 1`] = `
   type="warning"
 >
   <div>
-    The description below includes sections that have been automatically generated using an AI language model. Please exercise discretion when interpreting the information provided.
+    <React.Fragment>
+      Some of the fields of this entry
+       
+      <React.Fragment>
+        (including sections of the 
+        <b>
+          description
+        </b>
+        )
+      </React.Fragment>
+       
+    </React.Fragment>
+    have been automatically generated using an AI language model. Please exercise discretion when interpreting the information provided.
+  </div>
+  <div
+    style={
+      {
+        "display": "flex",
+        "gap": "2em",
+        "paddingTop": "0.4em",
+      }
+    }
+  >
+    <Memo(Connect(Link))
+      href="https://interpro-documentation.readthedocs.io/en/latest/llm_descriptions.html"
+      target="_blank"
+    >
+      <span
+        className="small icon icon-common icon-book"
+      />
+       
+      Read more on description generation
+    </Memo(Connect(Link))>
+    <Memo(Connect(Link))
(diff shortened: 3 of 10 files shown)


typhainepl:
I feel like having AI generated at the top of the entry can be misleading. For example, when I see such information, my interpretation is that the model has been generated using AI, which is not the case here.
I remember we had discussed adding a label for all the fields generated by AI (i.e. entry name, short name and description) instead.

PTHR10003: Entry AI generated reviewed - Description imported from interpro but not LLM
PTHR10005: Entry AI generated unreviewed - Description imported from interpro but not LLM
=> when will we encounter such cases? Is it when the name of the signature has been generated by AI, but not the description?

gustavo-salazar:
> I feel like having AI generated at the top of the entry can be misleading. For example, when I see such information, my interpretation is that the model has been generated using AI, which is not the case here. I remember we had discussed adding a label for all the fields generated by AI (i.e. entry name, short name and description) instead.

Correct we discussed such an option, but we didn't agree on it, as it might create too many badges. So the version here is the simplest implementation for it, but of course, it is open to discussion. I could implement an alternative version that has a label on each field, so we can compare.


> PTHR10003: Entry AI generated reviewed - Description imported from interpro but not LLM PTHR10005: Entry AI generated unreviewed - Description imported from interpro but not LLM => when will we encounter such cases? Is it when the name of the signature has been generated by AI, but not the description?

Yes, that sort of cases, I don't really know if they would actually appear in the database, but given the returned flags, it is a possibility:  for example on `PTHR10003` the flag that an entry is LLM is `true`, but that entry doesn't have a description, so it takes it from the InterPro entry that has been integrated to. And that entry can itself have a LLM generated description, which is not the case for `PTHR10003`, but it is for `PTHR10005`

typhainepl:
> Correct we discussed such an option, but we didn't agree on it, as it might create too many badges. So the version here is the simplest implementation for it, but of course, it is open to discussion. I could implement an alternative version that has a label on each field, so we can compare.

I agree that having too many badges isn't ideal. Since we already have a label for the descriptions, we could change the label at the top to `AI-generated name and short name` or something similar. What do you think?

>Yes, that sort of cases, I don't really know if they would actually appear in the database, but given the returned flags, it is a possibility: for example on PTHR10003 the flag that an entry is LLM is true, but that entry doesn't have a description, so it takes it from the InterPro entry that has been integrated to. And that entry can itself have a LLM generated description, which is not the case for PTHR10003, but it is for PTHR10005

I don't think we'll encounter such cases, but using the label suggested above will help clarify.

matthiasblum (commented):
Looks good.

Regarding informing users about AI-generated info, we could simplify by showing a generic warning message under the entry's name, such as:

image

> This entry contains information, such as name, short name, or description, that has been generated using an AI language model. Please exercise discretion when interpreting the information provided.

This message clearly informs that some of the _annotations_ have been generated using AI (and not the model itself).

If at least one of the AI-generated information is reviewed (name, short name, at least one block of the description), the message would mention that the information has not been reviewed:

> This entry includes information, such as name, short name, or descriptions, that has been generated using an AI language model and has not undergone review by expert curators. Please exercise discretion when interpreting the information provided.

I don't think that users will want to know what _exactly_ has been generated using AI (name and short name? which part of the description?), but if some want to, the information will be available through the REST API.

---

Also, the "Provide feedback" and "Add your annotation" are somehow redundant when looking at InterPro entries. We don't need to do the change as part of this PR, but eventually we could remove the link to the Google Form (and maybe change the text of "Add your annotation" to "Provide feedback").

---

> PTHR10003: Entry AI generated reviewed - Description imported from interpro but not LLM

Won't happen. We'll let curators flag an InterPro entry with AI-generated name and short name as reviewed, but not member database signatures (technically possible but complicated).

gustavo-salazar:
OK will implement a version with the suggestions. And will share here again for discussion.

---

@matthiasblum Some questions/comments from your feedback:

> I don't think that users will want to know what _exactly_ has been generated using AI (name and short name? which part of the description?), but if some want to, the information will be available through the REST API.

Currently, the API would identify if descriptions, or "entries" are LLM, not really identifying if it was the name, short_name, etc. Are you thinking of adding more details to the LLM flag?

> Also, the "Provide feedback" and "Add your annotation" are somehow redundant when looking at InterPro entries. We don't need to do the change as part of this PR, but eventually we could remove the link to the Google Form (and maybe change the text of "Add your annotation" to "Provide feedback").

I'll create a GH issue for this.

> > PTHR10003: Entry AI generated reviewed - Description imported from interpro but not LLM
>
> Won't happen. We'll let curators flag an InterPro entry with AI-generated name and short name as reviewed, but not member database signatures (technically possible but complicated).

I wasn't talking about AI generated signatures. My point is that when a member DB entry doesn't have a description, we import it from its interpro. (e.g. [PTHR10003](https://www.ebi.ac.uk/interpro/entry/panther/PTHR10003/)), but that description itself could have been LLM generated itself. Which I guess is useful to inform the user about it.

matthiasblum:
> > I don't think that users will want to know what _exactly_ has been generated using AI (name and short name? which part of the description?), but if some want to, the information will be available through the REST API.
> 
> Currently, the API would identify if descriptions, or "entries" are LLM, not really identifying if it was the name, short_name, etc. Are you thinking of adding more details to the LLM flag?

No, no. What I meant is that the API now has `is_llm` at the entry level, and `llm` at the description level. If `is_llm` is `true`, the name _and_ short name have been generated using AI. If `llm` is `true`, the description has been generated using AI. And if there is a mix of AI-generated and human-curated descriptions, it is possible to know the source/origin of each one (since there is one `llm` field per description). So if users want to know which annotation in the entry has been generated using AI, they can get this information from the API.

>> Won't happen. We'll let curators flag an InterPro entry with AI-generated name and short name as reviewed, but not member database signatures (technically possible but complicated).
>
> I wasn't talking about AI generated signatures. My point is that when a member DB entry doesn't have a description, we import it from its interpro. (e.g. [PTHR10003](https://www.ebi.ac.uk/interpro/entry/panther/PTHR10003/)), but that description itself could have been LLM generated itself. Which I guess is useful to inform the user about it.

At the moment, we use AI to generate annotations of unintegrated signatures, then we plan to auto-create InterPro entries by integrating these signatures and use the AI-generated annotations of the signature for the entry. So if an entry has a AI-generated description, it's because it has been imported from the signature. But your point stands as curators could integrate a signature (that doesn't have a description) in an entry which was originally created from another signature (that has an AI-generated description). :exploding_head:

gustavo-salazar:
Ok, I just did a minor version change where I removed the Badge for AI-generated Entries, and simplified the text in the callout. But given that we have the info on which field was LLM generated I added it as a Tooltip, that way the info is still there but doesn't take much space, and the message stays simple.

typhainepl (approved):
The new version looks better :)

src/components/Entry/BadgeAI/style.css
```diff
@@ -0,0 +1,21 @@
+.vf-badge--tertiary {
+  background: #e8e8e8;
```
matthiasblum, line 2:
```suggestion
  background: rgba(255, 220, 220, 0.5);
```

src/components/Entry/BadgeAI/style.css
```diff
@@ -0,0 +1,21 @@
...
+  border-color: transparent;
+  border-radius: 0.3em;
+  color: rgba(0, 0, 0, 0.6);
```
matthiasblum, line 5:
```suggestion
  color: rgba(159, 58, 56, 1);
```

src/components/Entry/BadgeAI/style.css
```diff
@@ -0,0 +1,21 @@
...
+  text-transform: none;
+  user-select: none;
+  &.mini {
+    font-size: 0.7rem;
+    padding: 2px;
+  }
+}
+.vf-badge--tertiary:hover,
+.vf-badge--tertiary:visited {
+  color: rgba(0, 0, 0, 0.6);
```
matthiasblum, line 16:
```suggestion
  color: rgba(159, 58, 56, 1);
```

PR #590, opened 2024-03-01
Minor issues

### LINKED ISSUE #482: Modifiers are not propagated to the download form
When browsing with a modifier, the **See more download options** (under **Export**) link does not include the modifier and its value.

### Example 1: browsing proteins in a given domain organization

https://www.ebi.ac.uk/interpro/protein/UniProt/?ida=2f2efb82f7c6f7394d8a5ef687352a767f74c585

The link to the download form (TSV format) is: https://www.ebi.ac.uk/interpro/result/download/#/protein/UniProt/|tsv
It should be: https://www.ebi.ac.uk/interpro/result/download/#/protein/UniProt/?ida=2f2efb82f7c6f7394d8a5ef687352a767f74c585|tsv

### Example 2: browsing InterPro entries having a specific GO term

https://www.ebi.ac.uk/interpro/entry/InterPro/?go_term=GO:0006355

The link to the download form (TSV format) is: https://www.ebi.ac.uk/interpro/result/download/#/entry/InterPro/|tsv
It should be: https://www.ebi.ac.uk/interpro/result/download/#/entry/InterPro/?go_term=GO:0006355|tsv

### LINKED ISSUE #564: conservation worker not working in production
Also, the `conservation` checkbox is working fine locally, but it is disabled in the future branch. Example: http://wp-np3-bd/future/entry/pfam/PF00002/entry_alignments/?type=seed

_Originally posted by @typhainepl in https://github.com/ProteinsWebTeam/interpro7-client/issues/562#issuecomment-1888763492_

### LINKED ISSUE #575: Add the option to open signature in a new tab from the clan schema
It would be nice to be able to open the families in a new tab when clicking on the graph found in clan pages. ATM we can only open it in the same tab.

### LINKED ISSUE #576: The split view doesn't work on Laptop screen
The split view doesn't work on Laptop screen, the protein viewer is not shown fully.
Example:
image

### LINKED ISSUE #577: Sunburst legends out of viewport
Reported on InterHelp:
> When I view taxonomy in a sunburst viewer, the control panel which shows the species name etc is now below the sunburst circle which means I can't view the name as I hover over a section of the sunburst circle. 

This is due to the new CSS layout used in it and to the fact that the sunburn component tries to resize itself on a greedy strategy. You can trigger this behaviour by resizing the screen when the sunburst is on display.

We need to make sure that the only time the labels go under is on small screens/viewports.

### LINKED ISSUE #578: Improve the styling of the switch
The `Switch` component used in the IDA search and other pages, needs some styling.
Especially when the labels of the switch are longer than the default on/off. For example in the similar proteins page.

### LINKED ISSUE #580: Some domain architectures height keeps growing
THere is a bug on certain IDA pages where the track displaying an entry keeps growing taking over the page. For example [PF00008](https://www.ebi.ac.uk/interpro/entry/pfam/PF00008/domain_architecture/)

gustavo-salazar:
A collection of small fixes reported on GH issues:

* FIX #580: There is a bug in nightingale where the `height` parameter of the interpro track is not doing anything. It has been reported, but on the interpro website we will avoid this issue by setting a `max-height` in the container div.
* FIX #575: Now it detects if the command or alt button are pressed while clicking on a Pfam entry on the clan visualization, and if so, opens the link in a new tab.
* FIX #564: The problem comes down to NIghtingale using a different syntax for web workers, as it is using rollup and not webpack to bundle the components. Webpack5 is supposed to support the same syntax, so I tried to move into that, but I didn't manage, and after many hours trying different setups, I gave up and changed the nightingale component using `patch-package`.
* FIX #576: Somehow the Mol* viewer is too greedy and in some cases, it was using more than 50% of the split view. I'm forcing it to stay in place via CSS `max-width: 50vw`.
* FIX #577: Gave both the sunburst and the labels the same wight in `flex-grow`; set a fix width for both (`70%` and `30%`), and change it to `100%` for small screens.
* FIX #578: Now `<ToggleSwitch>` has a prop `width` to make it wider when the labels are too long. Also played with CSS variables in the style to simplify the logic of the element size.
* FIX #482: The tooltip now receives the `search` object, which includes the modifiers, and includes it in the link to the Download Form. The Form was also updated to wait until it has received the OpenAPI payload, before refreshing the parameters.

### DIFF
--- a/patches/@nightingale-elements+nightingale-msa+4.5.0.patch
+++ b/patches/@nightingale-elements+nightingale-msa+4.5.0.patch
@@ -1,0 +1,23 @@
+diff --git a/node_modules/@nightingale-elements/nightingale-msa/src/nightingale-msa.ts b/node_modules/@nightingale-elements/nightingale-msa/src/nightingale-msa.ts
+index a2fb149..3f9b144 100644
+--- a/node_modules/@nightingale-elements/nightingale-msa/src/nightingale-msa.ts
++++ b/node_modules/@nightingale-elements/nightingale-msa/src/nightingale-msa.ts
+@@ -14,6 +14,7 @@ import NightingaleElement, {
+ } from "@nightingale-elements/nightingale-new-core";
+ import object2style from "./utils/object2style";
+ import { Region, SequencesMSA } from "./types/types";
++import ConservationWorker from "./workers/conservation.worker.ts";
+ 
+ const DEAFULT_TILE_HEIGHT = 20;
+ const DEAFULT_COLOR_SCHEME = "clustal2";
+@@ -53,9 +54,7 @@ class NightingaleMSA extends withManager(
+   })
+   overlayConservtion?: boolean = false;
+ 
+-  worker = new Worker(
+-    new URL("./workers/conservation.worker.ts", import.meta.url)
+-  );
++  worker= new ConservationWorker();
+   
+   private sequenceViewer?: SequenceViewerComponent | null;
+   private labelPanel?: LabelsComponent | null;
--- a/src/components/DownloadForm/URLParameters/index.js
+++ b/src/components/DownloadForm/URLParameters/index.js
@@ -119,11 +119,12 @@ SelectedParameter.propTypes = {
 const URLParameters = ({ type, data, search, onChange }) => {
   const [toAdd, setToAdd] = useState(null);
   const [toRemove, setToRemove] = useState(null);
+  const payloadIsReady = data && !data.loading && data.ok && data.payload;
   useEffect(() => {
     if (toAdd) setToAdd(null);
   });
   useEffect(() => {
-    return () => onChange({ target: null });
+    return () => payloadIsReady && onChange({ target: null });
   }, [toRemove]);
   const selectedParameters = Object.keys(search || {});
   if (toAdd) {
@@ -137,7 +138,7 @@ const URLParameters = ({ type, data, search, onChange }) => {
       setToRemove(null);
     }
   }
-  if (data.loading || !data.ok || !data.payload) return null;
+  if (!payloadIsReady) return null;
   const parameters = Object.fromEntries(
     (data.payload.paths?.[`/${type}/{sourceDB}`]?.get?.parameters || [])
       .filter((p) => p.$ref)
--- a/src/components/File/FileButton/index.tsx
+++ b/src/components/File/FileButton/index.tsx
@@ -39,6 +39,7 @@ export type FileButtonProps = {
   handleClick: (event: Event) => void;
   shouldLinkToResults?: boolean;
   showIcon?: boolean;
+  search?: Record<string, string>;
 };
 
 const FileButton = ({
@@ -56,6 +57,7 @@ const FileButton = ({
   shouldLinkToResults = true,
   showIcon,
   minWidth,
+  search,
 }: FileButtonProps) => {
   const downloading = Number.isFinite(progress) && !successful;
   const failed = successful === false;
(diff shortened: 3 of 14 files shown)


typhainepl:
FIX https://github.com/ProteinsWebTeam/interpro7-client/issues/564 is still not working for me locally
FIX https://github.com/ProteinsWebTeam/interpro7-client/issues/576 is still not working for me locally
FIX https://github.com/ProteinsWebTeam/interpro7-client/issues/577 Would it be possible to also have the 70/30 ratio for small screens (e.g. laptop)?
FIX https://github.com/ProteinsWebTeam/interpro7-client/issues/578 I'm not sure what is the impact of this change. Do you have a page example?

gustavo-salazar:
> FIX #564 is still not working for me locally 
> FIX #576 is still not working for me locally 

Damn! Those are cases of: it works on my laptop! 
On the one about conservation, maybe wait until it's deployed. as the issue was only happening on production builts.
ON the split-screen case, I'll play a bit more and see if I can replicate the issue again.

> FIX #577 Would it be possible to also have the 70/30 ratio for small screens (e.g. laptop)? 

Ehhh... I don't consider laptop a small screen. when I said small screens are mobile, maybe tablets. but in general on a laptop will depend on the size of the window.

> FIX #578 I'm not sure what is the impact of this change. Do you have a page example?
In https://www.ebi.ac.uk/interpro/protein/UniProt/P99999/similar_proteins/#table click the switch that is preselected with `UniProt KB` then you will see the issue that I'm trying to solve.

typhainepl:
> Ehhh... I don't consider laptop a small screen. when I said small screens are mobile, maybe tablets. but in general on a laptop will depend on the size of the window.

I asked because the sunburst view hadn't changed the first time I tested for some reason, now it's fine. 

> In https://www.ebi.ac.uk/interpro/protein/UniProt/P99999/similar_proteins/#table click the switch that is preselected with UniProt KB then you will see the issue that I'm trying to solve.
Looking good to me

Aside from #564 and #576 everything looks good to me.

matthiasblum (approved):
LGTM, only one minor suggestion. Feel free to merge. Thanks!

| Issue | Status |
|----------|----------|
| #580  | OK, does not expand anymore |
| #575  | OK, but should use the `Ctrl` key for Linux/Windows users, not `Alt` |
| #564  | Works locally |
| #576  | Still truncated on an old MacBook with a resolution of 1280x800. It seems that the 3D viewer wants to expand and doesn't stay nicely on its side |
| #577 | Looks good |
| #578 | Looks good (radio buttons >> switch, anyway) |
| #482 | Looks good|

gustavo-salazar:
> * [#576](https://github.com/ProteinsWebTeam/interpro7-client/issues/576): still truncated on an old MacBook with a resolution of 1280x800. It seems that the 3D viewer wants to expand and doesn't stay nicely on its side

I was going to blame my fancy new laptop, but I just reduced the resolution and I'm still not able to reproduce it.

matthiasblum:
> > * [#576](https://github.com/ProteinsWebTeam/interpro7-client/issues/576): still truncated on an old MacBook with a resolution of 1280x800. It seems that the 3D viewer wants to expand and doesn't stay nicely on its side
> 
> I was going to blame my fancy new laptop, but I just reduced the resolution and I'm still not able to reproduce it.

IMO, it's not a major issue. We can re-open the issue if we're contacted by users regarding this.

typhainepl:
#564 is now working locally
I found out that #576 doesn't work because my laptop resolution isn't the default, but 1352x878, it works fine when I set the resolution to the default, but I found a way around by zooming out to 80%.

typhainepl (approved):
Looks good to me

PR #591, opened 2024-03-04
Proteome page improvements

gustavo-salazar:
2 additions to the proteome page:
* An external link to Rfam: this one checks first the EBI search index of RFam to see if there is info there  for the given proteome
* A description directly fed from the UniProt API

### DIFF
--- a/src/components/Proteome/Summary/UniProtDescription/index.tsx
+++ b/src/components/Proteome/Summary/UniProtDescription/index.tsx
@@ -22,7 +22,6 @@ const getProteomeUrl = createSelector(
   (state: GlobalState) => state.customLocation.description.proteome.accession,
   (accession) => {
     if (!accession) return null;
-    console.log(config.root);
     return `${config.root.UniProt.href}/proteomes/${accession}`;
   },
 );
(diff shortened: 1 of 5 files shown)


src/components/Proteome/Summary/UniProtDescription/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import config from 'config';
+
+interface LoadedProps extends LoadDataProps<UniProtProteomesPayload> {}
+
+const UniProtDescription = ({ data }: LoadedProps) => {
+  if (!data) return null;
+  const { loading, payload } = data;
+  if (loading || !payload) return <Loading />;
+  return <div>{data.payload?.description}</div>;
```
matthiasblum, line 18: Handle cases where no description is available (e.g. UP000002473) and add a "Description" header, like we have for entries:

```suggestion
  return data.payload?.description ? (
    <>
      <h4>Description</h4>
      <p>{data.payload?.description}</p>
    </>
  ) : null;
```

matthiasblum (approved):
I've added a suggestion that should be included, but otherwise it looks good. Thanks!

src/components/Proteome/Summary/UniProtDescription/index.tsx
gustavo-salazar: Fair enough

PR #594, opened 2024-03-06
Ts DownloadForm

gustavo-salazar:
Migration to TypeScript and Visual Framework 2.0 for `DownloadForm` and all its sub-components.

Other changes:
* New component `<InputGroup>` to group this pattern and abstract its styling.
* Created a basic story for a `<InputGroup>` to test the styling works without foundation.


matthiasblum (commented):
Looks ok overall, but there are a few points that might need to be addressed

---

Why the `type` modifier is not a select anymore?

image

---

The progress bar seems to have disappeared, so after clicking "Generate" it's not possible to see if anything is happening. Is it expected and, if it isn't, can you reproduce?

---

While the download is running in the background, the date is a mix of `NaN` and `undefined`:

image

gustavo-salazar:
> Why the `type` modifier is not a select anymore?

Well spotted, it was just me introducing bugs. Fixed.

> The progress bar seems to have disappeared, so after clicking "Generate" it's not possible to see if anything is happening. Is it expected and, if it isn't, can you reproduce?

Yes, I can reproduce. But this one wasn't A bug I was introducing now. It was one that I introduced last year 😱
The `mapStateToProps` of the new `loadData` was only taking into account the new props (e.g. `data`, `isStale`) but not passing the props of the wrapped. But this is the first migrated component that makes use of it. Hence it was unnoticed until now. Fixed.

> 
> While the download is running in the background, the date is a mix of `NaN` and `undefined`:

Now this one is at least 2 years old, but probably more, It's likely it has always been like that. I just changed that to `"Currently running"` when the progress is less than 1.

matthiasblum (approved):
LGTM. Feel free to merge.

PR #606, opened 2024-04-22
Update/migrate footer

matthiasblum:
EMBL-EBI services now have to display a link to the [long-term data preservation](https://www.ebi.ac.uk/long-term-data-preservation).

And when you don't know where to put these kinds of corporate links, the footer is the perfect place to group them. To _bury_ them, you say? Maybe. But it's where it is on the [EMBL-EBI home page](https://www.ebi.ac.uk/) :shrug:.

This PR migrates the EBIFooter component to TS/VF so it looks like the footer on the EMBL-EBI home page.


src/components/EBIFooter/index.tsx
```diff
@@ -0,0 +1,296 @@
+import React from 'react';
+import Link from 'components/generic/Link';
+import cssBinder from 'styles/cssBinder';
+import local from './styles.css';
+const css = cssBinder(local);
+
+const EBIFooter = () => {
+  const year: string = new Date().getFullYear().toString();
+  return (
+    <div className="vf-footer">
```
gustavo-salazar: I'm adding in the CSS a bit of padding left and right so it doesn't get on the edge of the window when the window is small.

gustavo-salazar (approved):
I added a commit with a few minor style changes, including the changes for the inline comments.

BTW, not sure if you used it or not, but when doing the migration change the extension of the file with `git mv file.js file.tsx` that way the history of commits keeps attached to the new file. Unfortunately either git or github every now and then decides there are too many changes and end up with 2 different files. But when it works is nice!

src/components/ElixirFooter/__snapshots__/test.js.snap
```diff
@@ -2,76 +2,100 @@
 
 exports[`<ElixirFooter /> should render 1`] = `
 <div
-  className="row columns"
+  className="vf-u-fullbleed elixir-footer"
```
gustavo-salazar, line 5: Forgot to mention... I'm so glad of this change, as this fixes the case where the elixir header ends up over the filters, not only was ugly but also a bit of a pain on the puppeteer script.

matthiasblum:
> I added a commit with a few minor style changes, including the changes for the inline comments.

Cool. thanks!

> BTW, not sure if you used it or not, but when doing the migration change the extension of the file with `git mv file.js file.tsx` that way the history of commits keeps attached to the new file.

I _think_ I did, but now I have a doubt because the JS/TSX files are too big. :see_no_evil:

PR #607, opened 2024-04-24
Ts EBI header

gustavo-salazar:
Typescript and visual framework migration of EBIHeader.

There is a bit of redundancy on styles with the `EBI_global.css` that is still loaded in the website until we completely drop EBI framework 1.4.

The new story in storybook shows the header loading using only the `vf-` classes.

Also removing the EMBLDropdown component that hasn't been used for a while.


matthiasblum (changes requested):
Thank you.

The items of the header are all on the right-hand side of the page. I would expect them to be inside a container with a `max-width`:

`dev` branch:

image

`ts-ebi-header` branch:

image

---

Also, I am not sure _how_, but the changes seem to be responsible for style changes _outside_ the header.

`dev` branch:

client-dev

The **InterPro** header is rendered using Ubuntu Bold (on my system), due to `font-weight: 600`.

`ts-ebi-header` branch:

client-ts-ebiheader

Same header is now rendered with `font-weight: 400`. The main body/container of the page is a bit wider on `dev`. I believe both changes are due to Foundation's rules taking priority on  `ts-ebi-header`. But why or how, I have no idea.

gustavo-salazar:
OK, so the `global_ebi.css` includes redefinitions for high-level elements (e.g. `<body>`). On the migration to VF, we are not including that for the header anymore. I thought that wouldn't be a problem because it is included elsewhere. However, there are other files defining the same high-level objects(e.g. foundation).
And given that the CSS specificity of the files was the same (the selector in all of them is `body {`), it comes down to which one is the last one to be included. In the case of files, it comes down to what gets loaded first, making it a bit random, so sometimes the headers are defined by one file and other times by another. An unfortunate consequence of this transitional state between frameworks.

My solution to start detaching any dependency on `global_ebi.css` is to start defining these values in `interpro-vf.css` but also add a class to the `body` so the selector there has more specificity. 
This might have side-effects on particular styling of some headers(e.g. subtitle in main banner), and we will need to keep an eye for such cases

src/components/EBIHeader/style.css
```diff
@@ -1,25 +1,68 @@
...
-    float: right !important;
+        & > a {
+          &,
+          &:visited,
+          &:focus {
+            color: white;
+          }
+          &:not(.float-right):hover {
+            color: #373a36;
+            background-color: #d0d0ce !important;
```
matthiasblum, line 22: I don't know why I originally didn't use the custom properties.

```suggestion
            color: var(--vf-color--grey--darkest);
            background-color: var(--vf-color--grey--lightest);
```

src/components/EBIHeader/style.css
```diff
@@ -1,25 +1,68 @@
...
 /* Overwrite EBI guidelines styles… */
-.global-nav > * {
-  float: initial !important;
+.masthead-black-bar {
+  --vf-stack-margin--custom: 0;
+  background-color: var(--vf-color--grey--darkest);
+  /* border-top: 0 solid #eee; */
+
+  & > nav {
+    max-width: var(--vf-body-width, 80em);
```
matthiasblum, line 8: Center the `nav`:

```suggestion
    max-width: var(--vf-body-width, 80em);
    margin: 0 auto;
```

matthiasblum (changes requested):
Thanks for finding the issue.

I made two small suggestions to the CSS file, and there are a few lines that can be removed in `index.template.html`:

 ```css
      .tmp-ebi-header {
        background-color: #373a36;
        height: 37.289px;
      }
```

gustavo-salazar:
> I made two small suggestions to the CSS file, 

Cool, merged them.

> and there are a few lines that can be removed in `index.template.html`:
> 
> ```css
>      .tmp-ebi-header {
>        background-color: #373a36;
>        height: 37.289px;
>      }
> ```

This actually has a purpose: the idea is that it uses this minimal CSS to create the block for the banner and show something while all the react components get loaded, so there is a minimal skeleton of the page instead of a blank page.

matthiasblum:
> This actually has a purpose: the idea is that it uses this minimal CSS to create the block for the banner and show something while all the react components get loaded, so there is a minimal skeleton of the page instead of a blank page.

I see. Then maybe the `height` rule could be removed? `37.289px` is more than the max height of nav bar items.

matthiasblum (approved):

gustavo-salazar:
> I see. Then maybe the `height` rule could be removed? `37.289px` is more than the max height of nav bar items.

Nope. It is exactly the height that will be displayed. So when just the HTML gets loaded we got this:
Screenshot 2024-05-01 at 09 58 27

And when React takes over we get this:

Screenshot 2024-05-01 at 09 58 00

PR #608, opened 2024-04-24
EBI survey banner in home page

gustavo-salazar:
Adding the banner details as a Callout on the Home page.

**NOTE:** Only to be merged the 20th May

### DIFF
--- a/src/pages/Home/index.js
+++ b/src/pages/Home/index.js
@@ -12,12 +12,11 @@ import Tabs from 'components/Tabs';
 import Description from 'components/Description';
 import CurrentVersion from 'components/home/CurrentVersion';
 import GeneralWarning from 'components/home/GeneralWarning';
+import EBISurvey from 'components/home/EBISurvey';
 import InterProGraphicAnim from 'components/home/InterProGraphicAnim';
 import { PrintedInterPro2022 } from 'components/Help/Publication';
 import Tip from 'components/Tip';
 import Toast from 'components/Toast/Toast';
-// $FlowFixMe
-import Callout from 'components/SimpleCommonComponents/Callout';
 
 import Link from 'components/generic/Link';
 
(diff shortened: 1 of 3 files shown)


src/pages/Home/index.js
```diff
@@ -252,6 +254,23 @@ class Home extends PureComponent {
           />
         ) : null}
         <GeneralWarning />
+        <Callout type="warning">
+          <div>
+            Do data resources managed by EMBL-EBI and our collaborators make a
+            difference to your work?
+            <br /> Please take 10 minutes to fill in our{' '}
+            <Link
+              href="https://www.surveymonkey.com/r/HJKYKTT?channel=[webpage]"
```
matthiasblum, line 263: I suspect the `channel` param will be used to track the resources users were visiting when opening the survey.

```suggestion
              href="https://www.surveymonkey.com/r/HJKYKTT?channel=InterPro"
```

matthiasblum (approved):
LGTM.

If it's possible, it would be nice to have the URL of the survey in the config file, and only display the callout if the URL is not undefined or an empty string. This way, we can enable/disable the callout without having to change the code (just update the config file, build, and deploy).

src/pages/Home/index.js
gustavo-salazar: I initially thought the same, but in the document the link changes in other parts as `channel[email]`, `channel[twitter]`, etc. so I leave it as `[website]`

matthiasblum: I noticed that too, but I thought the links be updated, e.g.:

- link in an email: `channel=recipient@company.com`
- link in a X/Twitter post: `channel=@interprodb`
- link in a LinkedIn post: `channel=interpro-pfam`

I'll reach the people in charge of the survey and ask them to clarify.

gustavo-salazar: cool, let's wait for that...  but in the meantime, I implemented your suggestion about the link in the config. Please have a look to it

matthiasblum (approved):
LGTM

src/pages/Home/index.js
matthiasblum: Got confirmation that indeed we should leave it as it is.

PR #609, opened 2024-05-01
Interpro Button

### LINKED ISSUE #566: Use a `<Button>` component
To reduce a bit the different styles of buttons we can create a component similar to [the one in storybook](https://github.com/ProteinsWebTeam/interpro7-client/blob/storybook-again/stories/Button.tsx) and then use it around our code instead of direct use of `<button>`

### LINKED ISSUE #581: Styling the Export dropdown content
The content of the export dropdown button needs some attention.
For example, the `Web View` option uses the old foundation style for buttons.

gustavo-salazar:
A simple component to wrap the usage of `<button>` elements so they all share the same styling base. Developed in Typescript with the new visual framework, but adding a CSS customization to it.

Most of the files are just straight-out replacements of `<button>` by `<Button>` both in `.js` and `.tsx` files

FIX #566 
FIX #581


src/pages/Settings/index.js
```diff
@@ -867,6 +864,8 @@ class Settings extends PureComponent /*:: <SettingsProps, SettingsState> */ {
               </nav>
               <section>
                 <section onChange={changeSettings}>
+                  <AddToHomeScreen />
+                  <hr />
```
matthiasblum, line 868:
```suggestion
```

First time I have a look at this component, since it doesn't render on my system (tried Firefox and Chromium). It's unclear to me how this is useful, especially on the settings page.

src/subPages/EntryAlignments/index.js
```diff
@@ -59,9 +62,7 @@ const Alignment = ({
           </p>
           <p>
             If you still want to display it, press{' '}
```
matthiasblum, line 64:
```suggestion
```

src/subPages/EntryAlignments/index.js
```diff
@@ -59,9 +62,7 @@ const Alignment = ({
...
-            <button className={f('button')} onClick={() => setForceShow(true)}>
-              HERE
-            </button>
+            <Button onClick={() => setForceShow(true)}>HERE</Button>
```
matthiasblum, line 65:
```suggestion
            <Button onClick={() => setForceShow(true)}>Open alignment</Button>
```

matthiasblum (changes requested):
Looks good.

I made a few suggestions, and there are a few cases where some styling could be added:

- Home page
  - Buttons in tab components (e.g. Latest entries, species, etc.) do not use the `Button` component (or its style)
  - the `Read all articles` uses the 3D button
- Protein summary page
  - The `Download matches (TSV)` and `Download sequence (FASTA)` buttons are stuck together
  - The `Search with InterProScan` and `Download sequence (FASTA)` use the  3D button
- Search by domain architecture page: the `InterPro/Pfam` toggle and `Export` button are no longer aligned
- Browse by proteome page: the text inside the "Generate" buttons looks right-aligned, possibly because the column is narrower due to the name of proteomes being long. Maybe a padding could be applied.

src/pages/Settings/index.js
gustavo-salazar: This is for PWA, with it you can add an icon of InterPro as an app in your mobile, see https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeinstallprompt_event

I do use it that way, but I wonder how many people have also installed like that. Maybe, having it in settings is hiding it a bit too much.

matthiasblum: Yes, I ended up on this page as well. Do we know how many users access InterPro from a smartphone or tablet? Because if nearly all users access InterPro using a web browser on a desktop/laptop, I'd say this component can be removed.

gustavo-salazar: 14K in the last month, It is quite significant!

Screenshot 2024-05-10 at 12 10 31

matthiasblum: OK, then this is something that we can keep. Not the scope of this PR, but maybe the component could be displayed somewhere with more visibility than the settings page (side menu?).

matthiasblum (approved):
Looks good, thanks!

PR #605, opened 2024-04-22
Links to Foldseek

matthiasblum:
Add links to Foldseek so users can perform protein structure searches. Links are added on the summary pages of PDBe structures (e.g. `/interpro/structure/PDB/10gs/`) and protein pages (if an AlphaFold prediction is available, e.g. `/interpro/protein/unreviewed/A0A009GN48/`).

PR #604, opened 2024-04-22
Display short names

matthiasblum:
Add a column showing the short name of the entry/signature, e.g.

* InterPro entries: `/interpro/entry/InterPro/#table`
* Pfam signatures: `/interpro/entry/pfam/#table`
* NCBIfam signatures matching `A0A000`: `/interpro/protein/UniProt/A0A000/entry/ncbifam/#table`
* PIRSF signatures found in the _Hibiscus syriacus_ proteome: `/interpro/proteome/uniprot/UP000436088/entry/pirsf/#table`

PR #611, opened 2024-05-13
Dev => master Release `99.1`

gustavo-salazar:
Most(maybe all) of the code in this PR has already been reviewed in: 

- #604 
- #605 
- #606 
- #607 
- #608 

So the purpose of it is to check it all together before merging.


matthiasblum (approved):
Looks good to me.

Regarding the header, I realize that I might not have explained well enough why I wanted to remove the fixed `height` in #607 ([link to message](https://github.com/ProteinsWebTeam/interpro7-client/pull/607#issuecomment-2088056435)).

[As you mentioned](https://github.com/ProteinsWebTeam/interpro7-client/pull/607#issuecomment-2088179975), the header of the `<header>` element is `37.289px` after the page is loaded:

interpro-screenshot-1

However, the inner `<nav>` and `<ul.nav>` elements don't need "that much":

interpro-screenshot-2

which leads to this when hovering the items:

interpro-screenshot-3

Removing the `height` rule in `index.template.html` shrinks the `<header>` element to the height of the `<nav>` element:

interpro-screenshot-4

typhainepl (approved):
Looking good to me :)

PR #613, opened 2024-05-16
Multiple wikipedia pages

gustavo-salazar:
The website now expects the Wikipedia field to be an array of objects.
And with it, it organizes the different articles in tabs at the end of the entry summary page.

Example PF00001


matthiasblum (approved):
Looks good to me, thanks.

I would just suggest to add a title to the Wikipedia section.

And also, in the `Wikipedia` component:

https://github.com/ProteinsWebTeam/interpro7-client/blob/d7a2c4829c6aad77d67dd44533564b3bbc663b80/src/components/Entry/Summary/Wikipedia/index.tsx#L83-L92

- switch the Wikipedia article title from `<h4>` to `<h5>`
- remove the `tag` element, which is not really necessary anymore (if there is the section has "Wikipedia" for title)

PR #617, opened 2024-05-31
wikipedia links in caption

gustavo-salazar:
The text for the caption in wikipedia panels might have links to wiki articles with the syntax `[[ link ]]`.

This PR parses those to create the links to the right place in wikipedia.

### DIFF
--- a/src/components/Entry/Summary/Wikipedia/index.tsx
+++ b/src/components/Entry/Summary/Wikipedia/index.tsx
@@ -17,6 +17,8 @@ import ipro from 'styles/interpro-vf.css';
 
 const css = cssBinder(local, ipro);
 
+const wikiLinkPrefix = 'https://en.wikipedia.org/wiki/';
+
 interface WikipediaProps
   extends WikipediaEntry,
     LoadDataProps<WikipediaPayload> {}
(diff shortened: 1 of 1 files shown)


src/components/Entry/Summary/Wikipedia/index.tsx
```diff
@@ -113,7 +115,24 @@ const Wikipedia = ({ title, extract, thumbnail, data }: WikipediaProps) => {
...
+                            href={`${wikiLinkPrefix}${text.replaceAll(
+                              /[[\]]/g,
+                              '',
+                            )}`}
+                          >
+                            {text.replaceAll(/[[\]]/g, '')}
+                          </Link>
+                        ) : (
+                          <span key={i}>{text}</span>
+                        ),
```
matthiasblum, line 133: Do we really need to wrap each piece of text in a `<span>` element?

```suggestion
                        ) : text,
```

matthiasblum (approved):
LGTM. I only made a minor suggestion.

src/components/Entry/Summary/Wikipedia/index.tsx
gustavo-salazar: This is more of a React requirement. When using a map to generate a list of elements, the element should have a key, which is important for dealing with re-renders.
That said, I forgot to put the Key in the `Link` element above. 🤦🏽

PR #618, opened 2024-06-03
Ipscan infomessages

gustavo-salazar:
Tidying up the warning messages for the Search by Sequence area.

Replacing the example sequence for human p53


src/components/IPScan/Search/SequenceInput/index.tsx
matthiasblum, line 1: Shouldn't it be

```ts
    const editorRef = useRef<Editor|null>(null);
```

instead of

https://github.com/ProteinsWebTeam/interpro7-client/blob/66225f95fc0958c2b4e28c888cca04dcde1f93ff/src/components/IPScan/Search/SequenceInput/index.tsx#L264

gustavo-salazar: There is an overload of the types definition of `useRef` that does exactly that:
```ts
function useRef<T>(initialValue: T | null): RefObject<T>;
```

src/components/IPScan/Search/SequenceInput/InfoMessages/index.tsx
```diff
@@ -1,64 +1,82 @@
...
+        <ul className={css('warnings')}>
+          {!validCharacters.result && (
+            <li>
+              The sequence with the header below has invalid characters. '
+              <div className={css('header')}>{validCharacters.header}</div>
+            </li>
+          )}
+          {tooShort.result && (
+            <li>
+              The sequence {tooShort.header ? 'with the header below' : ''} is
```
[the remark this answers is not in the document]
gustavo-salazar: After much tought I think you are right, and this is unnecessary. Will remove

gustavo-salazar:
> I made a few comments/suggestions. Also, is there a reason why each sequence is checked several times in different functions (`isTooShort()`, `hasDuplicateHeaders()`, etc.) and only the first sequence with an error is reported? Wouldn't it better to keep track of all errors?

The only real reason is about the way this code evolved. But considering that we can now talk about 100 sequences, it makes sense to update this code to make a single pass for performance's sake. 
I'll change it and ask for a re-review

matthiasblum (approved):
Looks good to me.

PR #621, opened 2024-06-06
Latest entries filter

gustavo-salazar:
Adds a new filter to the Browse By InterPro area, that allows filtering by the latest entries reported in the more recent release notes.


matthiasblum (changes requested):
Looks good overall. Thanks.

One thing I don't quite like is that the button/labels/switches should be "New entries" instead of "Latest entries". "New Entries" is the most intuitive and clear label for users, as it directly let them know that these are new, i.e. made public in the most recent release. In contrast, "Latest Entries" might suggest only the very last few additions.
If it's possible to change the labels, please do so. The API modifier can be left unchanged so there are no breaking changes/links.

matthiasblum (approved):
LGTM

PR #624, opened 2024-06-12
Ts filters

gustavo-salazar:
Typescript and Visual framework migration for the browse page filters.

* `EntryListFilters/*`
* `ProteinListFilters/*`
* `StructureListFilters/*`
* `FiltersPanel/*`
* `EntryListFilters/*`
* `MemberDBSelector`


tgrego (approved):
Looks good, everything seems to work fine.
The color bubble around the db size on the db selector used to have a padding that is now gone ( bubble covers exactly the text now). It's just visual and works as expected but maybe looked better before... Tried to change it back but without success!

tgrego:
Visually as before now, thanks!

src/components/Protein/ProteinListFilters/TaxonomyFilter/TaxonomyOption/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <NumberComponent
+        label
+        loading={loading}
+        className={css('filter-label')}
+        abbr
+      >
+        {count}
+      </NumberComponent>
+    )}
+  </label>
```
matthiasblum, line 52: When hovering the options, there is an animation on the counter instead of the label, unlike other filters.

gustavo-salazar: good catch! The css rule for it requires the text of the option(line 41) to be in a `<span>` tag. 
Fix included in the latest commit

gustavo-salazar:
> One thing I noted is that the filters panel is narrower on the structures page. This is not introduced by this PR (already on the live website), but causes longer option labels (e.g. CATH-Gene3D, Cryo-EM) to wrap. Can the panel have a minimum width?

I made the text for the member DB name `white-space: nowrap;` this way I won't have to guess a minimum width and the panel won't ever be narrower than the longest member DB name.

PR #626, opened 2024-06-18
Sortable table

### LINKED ISSUE #623: Permanent 408 (Query timeout) when browsing sets/clans
@blazaropinto reported that changing the number of results per page (to 50 or 100) when [browsing Pfam clans](https://www.ebi.ac.uk/interpro/set/all/entry/pfam/) leads to a query timeout that is never replaced by the results.

I noticed that the API request is:

```
https://www.ebi.ac.uk/interpro/wwwapi/set/all/entry/pfam/?page_size=50&extra_fields=counters:entry,description
```

When using the default page size, the payload is quite big. It is due to each clan having its members nested in the `entry_subset` property.

Is there are reason not to use the query below?

```
https://www.ebi.ac.uk/interpro/wwwapi/set/pfam/?page_size=20&extra_fields=counters,description
```

I think all the information needed is there.

gustavo-salazar:
The main  feature added here is sortable buttons for accession in:
* Browse by InterPro
* Browse by MemberDB
* Browse by Set

Changes included to achieve the objective:
* Add the `isSortable` prop to the `Table/Column` and update the logic to render the headers to split the logic that was already there for `isSearchable` in `Table/Header`
* Moving some table related css rules into `Table/styles.css`
* Migrating `Table/Header` into TS.
* In `loadData/deafults` we replace any request with `/set/[db]/entry/[db]` for `/set/[db]` which FIX #623 and allows the sorting by to be run in MySQL.

**Note 1:** This requires an [update in the API](https://github.com/ProteinsWebTeam/interpro7-api/commit/a2f710460d34c39ba3f3407d820d6a97410f3e4d) that is currently in the `dev` branch.
**Note 2:** First of a series of PRs that are chained to migrate `<Table>` and all its subcomponents. So when merge, delete the branch to update the next PRs.


tgrego (approved):
No issues, tables can be sorted by accession now.
Looks ready to go.

matthiasblum (approved):
Looks good to me.

Two minor suggestions:
- The sortable caret icons are traditionally placed on the right of the column name.
- The onhover/onclick events should be applied to the column name as well, not just the caret icons. If it's easier, it could applied to the entire header cell.

gustavo-salazar:
Changes requested done

matthiasblum:
Looks great, thanks! Feel free to merge.

PR #635, opened 2024-06-25
Additional links in interactions page

matthiasblum:
This PR adds links to IntAct and EuropePMC for the IntAct entries listed in the _interactions_ subpage of InterPro entries.


gustavo-salazar:
Mhh.. all seems OK, except that IntAct seems to be down, so I'm not able to check the links!
I will try later

PR #638, opened 2024-07-01
Get representative boolean from location instead of fragment

matthiasblum:
The first representative domain selection algorithm was evaluating fragments independently, i.e. one fragment from one domain could be selected as representative, but not the other fragment(s).

I changed that, and for the past few months, the `representative` boolean has been available at both the `entry_protein_locations` and `fragments` levels, e.g. [A0A023HJ61](https://www.ebi.ac.uk/interpro/api/entry/all/protein/UniProt/A0A023HJ61/).

This PR changes the source of `representative` to use the one at the location level.

The `representative` property will stopped being added at the fragment level for the next release.


gustavo-salazar (approved):
All good, happy for it to be merged

Issue #639, opened 2024-07-10
Release note history isn't in correct order


typhainepl:
The order of the release notes history is incorrect, as InterPro 100.0 is now at the top instead of being at the bottom of the list.

PR #640, opened 2024-07-17
Sort InterPro releases by version number

matthiasblum:
Addresses the issue reported in #639 (to reproduce, head to https://www.ebi.ac.uk/interpro/release_notes/ and click "Show Previous Releases).


gustavo-salazar (approved):
Not sure why I had the impression some releases had characters on them, e.g. `76.0b` but checking in the current table, it doesn't seem to be the case, so yeah ordering this way makes total sense

PR #642, opened 2024-07-18
Warning for IPScan pages update

gustavo-salazar:
Adds a warning about the upcoming version of ipscan results. Informing the users that old results will be incompatible and encourage them to save their jobs.


tgrego (approved):
Looks sensible to me

Issue #634, opened 2024-06-24
Invalid link to API download form in the documentation


matthiasblum:
The [Help page](https://www.ebi.ac.uk/interpro/help/documentation/) as a link to the [Code snippet generator](https://www.ebi.ac.uk/interpro/result/download/#/entry/interpro).

However, an error is raised when opening the page because the output format is missing. It should be `https://www.ebi.ac.uk/interpro/result/download/#/entry/interpro|tsv`, or (even better), the page should accept a missing format parameter and default to `tsv`.

Issue #633, opened 2024-06-24
HMM logo settings are displayed at the top of the page 


matthiasblum:
When looking the HMM logo of Pfam families, the settings of the viewer are not on the viewer itself, or just above, but at the top of the page.

Example: https://www.ebi.ac.uk/interpro/entry/pfam/PF00001/logo/

![image](https://github.com/ProteinsWebTeam/interpro7-client/assets/2097501/49c4670f-72ee-4d33-9b22-4184aecb4176)

PR #632, opened 2024-06-21
Ts table tree

gustavo-salazar:
Migration to typescript and visual framework for  `Table/view/Tree`
* Extracting `Table/view/Tree/DataProvider` 

The `pages/Taxonomy` was also migrated, so you can use the Browse By Taxonomy to test the view.

Other changes:
* Extracting `Taxonomy/List` from `pages/Taxonomy`
* Extracting `Taxonomy/ExactMatchSearch` from `pages/Taxonomy`

**Note:** Seventh and last of a series of PRs chained to migrate `<Table>` and all its subcomponents. Make sure to review #631 first.

PR #631, opened 2024-06-19
Ts table minor subcomponents

gustavo-salazar:
Migration to typescript and visual framework for minor subcomponents of `Table`

Components Migrated:
* `Table/Exporter`
* `Table/HighlightToggler`
* `Table/PageSizeSelector`
* `Table/SearchBox`
* `Table/TotalNB`

Any of the Browse By areas can be used for testing, however, the only one migrated to TS is Browse By Set.

Other changes:
* Fixing the page header indicator. issue reported in #622 
* Extracted the function toCanonicalURL to its own file to migrate it independently.

**Note:** Sixth of a series of PRs chained to migrate `<Table>` and all its subcomponents. Make sure to review #630 first. And, when merging, delete the branch to update the next PRs.

PR #643, opened 2024-07-18
InterPro Release 101

gustavo-salazar:
To be released together with InterPro 101.0 - Currently deployed in the Dev instance: http://wp-np3-bd/interpro/

### Pull Requests
Most if not all have been reviewed in the following PRs:

* #617 
* #618 
* #619 
* #620
* #621 
* #624 
* #626 
* #627 
* #628 
* #629 
* #630 
* #631 
* #632 
* #635 
* #638
* #640 

### Issues Fixed (Some not in PRs)
* #622 
* #623 
* #633
* #634


tgrego (commented):
I could replicate the wikipedia issues.
Otherwise looking good!

tgrego (approved):

gustavo-salazar:
Thanks Typhaine

> Looking good overall, but I've got a few comments: At first, when I wanted to test the changes made to the Wikipedia section, the page was showing an error message and wasn't displaying the Wikipedia section, even after hard refresh, but after waiting sometime and accessing the page, it is now working for PF00001, but not for PF00003.

Cool. Found a bug and will puch a fix in a minute

> Since we now have a "Feedback for entries button", I believe we could remove the "Provide feedback" from the warning message on AI-generated entries.

The Feedback button is only visible for InterPro and Pfam, but there are AI descriptions in other member DBs, so I think the link should stay in the message.

> There are some differences in the column names in tables throughout the website. I.e. some tables the column name in uppercase, other in lower case (e.g. pathways subpage) and others have some columns with upper case for the first letter, but other not (e.g. http://wp-np3-bd/interpro/entry/pfam/#table). It would be good to harmonise everything. I would suggest having only the first letter in upper case only. If the column is made of multiple words, the upper case should only be for the first word.

OK I put a CSS rule to capitalize all table headers, however `capitalize` also makes the first letter of second words uppercase, so not exactly how you wanted, but atlease this way all are harmonized. If you find one that is not please let me know. I'll deploy in wwwdev today so we can check the final version before release on Thursday

typhainepl (commented):
Looking good to me, you can merge the branch.

PR #645, opened 2024-07-25
Removed description from EntryCard and changed husky configuration

apolignano:
Removed the description and its related logic (as discussed with @gustavo-salazar). 

The result comes dows to this:
Screenshot 2024-07-25 at 12 44 20

The fix for the husky pre-commit file is to prevent running into the `command not found` error for lint-staged (@gustavo-salazar suggests that it might happen if `./node_modules/.bin/ ` is not added to PATH).

### DIFF
--- a/src/components/Entry/Card/index.tsx
+++ b/src/components/Entry/Card/index.tsx
@@ -19,16 +19,24 @@ type Props = {
     extra_fields?: {
       counters: MetadataCounters;
       literature?: Record<string, Reference>;
+      short_name: string;
     };
   };
   search: string;
   entryDB: MemberDB | 'interpro';
 };
 const EntryCard = ({ data, search, entryDB }: Props) => {
+  console.log(data.extra_fields);
   const name =
+    typeof data.extra_fields?.short_name === 'string'
+      ? data.extra_fields?.short_name
+      : data.metadata.accession;
+
+  const subtitleFullName =
     typeof data.metadata.name === 'string'
       ? data.metadata.name
       : data.metadata.name.name;
+
   return (
     <Card
       imageComponent={
@@ -43,7 +51,6 @@ const EntryCard = ({ data, search, entryDB }: Props) => {
         ) : (
           <Tooltip title={`${entryDB} database`}>
             <MemberSymbol
-              // size="2em"
               type={entryDB}
               aria-label="Database type"
               className={css('md-small')}
@@ -66,6 +73,7 @@ const EntryCard = ({ data, search, entryDB }: Props) => {
           <HighlightedText text={name || ''} textToHighlight={search} />
         </Link>
       }
+      subHeader={<>{subtitleFullName}</>}
       footer={
         <>
           {entryDB.toLowerCase() === 'interpro' ? (
--- a/src/components/Entry/Card/index.tsx
+++ b/src/components/Entry/Card/index.tsx
@@ -26,7 +26,6 @@ type Props = {
   entryDB: MemberDB | 'interpro';
 };
 const EntryCard = ({ data, search, entryDB }: Props) => {
-  console.log(data.extra_fields);
   const name =
     typeof data.extra_fields?.short_name === 'string'
       ? data.extra_fields?.short_name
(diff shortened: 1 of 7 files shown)


src/components/Entry/Card/index.tsx
```diff
@@ -146,21 +114,8 @@ const EntryCard = ({
         ) : (
           <Loading />
         )}
-        {showDescription &&
-          (desc && included ? (
-            <div className={css('new-card-description')}>
```
gustavo-salazar, line 151: Remove the bit of styling in the `style.css` file for the class `new-card-description`

gustavo-salazar (approved):
All good. Just a minor comment to avoid keeping CSS code that is not used.

I added @matthiasblum as a reviewer as well. As he suggested the change he might want to add something to it.

src/components/Entry/Card/index.tsx
```diff
@@ -14,53 +13,30 @@ import cssBinder from 'styles/cssBinder';
...
-  showDescription?: boolean;
 };
-const EntryCard = ({
-  data,
-  search,
-  entryDB,
-  showDescription = true,
-}: Props) => {
+const EntryCard = ({ data, search, entryDB }: Props) => {
+  console.log(data.extra_fields);
```
matthiasblum, line 29: Remove call to `console.log()`:

```suggestion
```

src/components/Entry/Card/index.tsx
```diff
@@ -14,53 +13,30 @@ import cssBinder from 'styles/cssBinder';
...
-        .sort(
-          (a, b) =>
-            (desc as StructuredDescription).text.indexOf(a[0]) -
-            (desc as StructuredDescription).text.indexOf(b[0]),
-        );
-    }
-  }
   return (
     <Card
       imageComponent={
```
matthiasblum, line 42: The `MemberSymbol` component isn't really informative, as it's simply a coloured database icon. Member database signatures also have a type, so I would prefer if 

https://github.com/ProteinsWebTeam/interpro7-client/blob/56637c3ea157cd10f82d374bc385629b97c24e2c/src/components/Entry/Card/index.tsx#L43-L59

was replaced by 

```ts
        <Tooltip title={`${data.metadata.type.replace('_', ' ')} type`}>
          <interpro-type
            dimension="2em"
            type={data.metadata.type.replace('_', ' ')}
            aria-label="Entry type"
          />
        </Tooltip>
```

src/components/Entry/Card/index.tsx
```diff
@@ -98,6 +73,7 @@ const EntryCard = ({
           <HighlightedText text={name || ''} textToHighlight={search} />
         </Link>
       }
+      subHeader={<>{subtitleFullName}</>}
       footer={
         <>
           {entryDB.toLowerCase() === 'interpro' ? (
```
matthiasblum, line 79: I don't think we need to display the entry type for InterPro entries, since the icon in the header represents the type already.

Instead the accession can be displayed on the left (for both InterPro entries and member database signatures). And the integration status can be displayed on the right (member database signatures only).

src/components/SimpleCommonComponents/Card/CounterIcon/styles.css
matthiasblum, line 1: I don't like that icons are underlined on `hover`.

Replace

https://github.com/ProteinsWebTeam/interpro7-client/blob/56637c3ea157cd10f82d374bc385629b97c24e2c/src/components/SimpleCommonComponents/Card/CounterIcon/styles.css#L16-L20

by

```ts
  & a:focus,
  & a:visited,
  & a:hover {
    color: #333;
    text-decoration: none;
  }
```

(This also fix the wrong pseudo-class, it's `:focus`, not `:focused`).

matthiasblum (changes requested):
Looks good overall. There are a few minor changes that I'd like to request (see individual comments).

Also, some short names are too long, and do not include hyphens, so they do not wrap and overflow, e.g.

image

Adding the `word-break: break-all;` rule to the card title styles should fix that:

https://github.com/ProteinsWebTeam/interpro7-client/blob/56637c3ea157cd10f82d374bc385629b97c24e2c/src/components/SimpleCommonComponents/Card/styles.css#L37-L42

apolignano:
- Now the title wraps and doesn't overflow;
- Accession number is on the left for and type gone for InterPro Entry cards
- DB member icon gone and type icon displayed for member DBs entries
- "Integrated into" shown on the right for member DBs entries
- No underlying when hovering counter icons

image
image

gustavo-salazar:
I'm happy with these changes!

matthiasblum (approved):
Looks great, thanks!
Feel free to merge.

gustavo-salazar:
Hey @apolignano  just noticed a couple of eslint warnings introduced in this PR. I guess to fix them you need to do it directly in dev as this has already been merged.

apolignano:
See https://github.com/ProteinsWebTeam/interpro7-client/commit/9398226d6e58638235c82ebc813dc978a982047c for the eslint warnings.

PR #648, opened 2024-07-31
Moved reference section below description

apolignano:
Changed the position of references as requested in [IBU-11215](https://www.ebi.ac.uk/panda/jira/browse/IBU-11215).

- The references are now above the GO Terms (e.g. IPR 000003)
image

- The references are now in the same section of the description, so that if the right section is too long, they don't get pushed down (e.g. IPR000001)

image


apolignano:
Tests were not passing because of the Jest snapshot, now updated. Tests passing now.

gustavo-salazar (approved):
Looks good to me!

matthiasblum (approved):
Looks good to me, thanks!

PR #649, opened 2024-07-31
Fixed matches expanding column

### LINKED ISSUE #637: Some tracks keep expanding horizontally
Does not affect all pages.

Example: [PDB structures matching Pfam entry PF11133](https://www.ebi.ac.uk/interpro/entry/pfam/PF11133/structure/PDB/#table)

Maybe related to #580 but for width instead of height?

apolignano:
The column displaying tracks kept expanding for some entries. max-width was required

Closes #637. The example reported was about entry [PF11133](https://www.ebi.ac.uk/interpro/entry/pfam/PF11133/structure/PDB/#table).

image

### DIFF
--- a/src/components/Matches/EntriesOnProtein/__snapshots__/test.js.snap
+++ b/src/components/Matches/EntriesOnProtein/__snapshots__/test.js.snap
@@ -44,6 +44,7 @@ exports[`<EntriesOnProtein /> should render 1`] = `
         margin-color="#fafafa"
         shape="roundRectangle"
         use-ctrl-to-zoom={true}
+        width={300}
       />
     </TooltipForTrack>
   </div>
--- a/src/components/Matches/EntriesOnProtein/index.tsx
+++ b/src/components/Matches/EntriesOnProtein/index.tsx
@@ -78,6 +78,7 @@ const EntriesOnProtein = ({ matches, match }: Props) => {
             id={`track_${entry.accession}`}
             data={data || []}
             shape="roundRectangle"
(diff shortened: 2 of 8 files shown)


matthiasblum (commented):
Fix doesn't seem to be working, at least on my end.

image

A few seconds later:

image

However, it does eventually stop expanding when the SVG containers reach a width of 400px.

Tested with Firefox and Chrome on Ubuntu and MacOS.

apolignano:
Ok, seems like it gives different results with different screen sizes and hence, window sizes. I had tested it on my laptop my now it keeps expanding when I open it on my monitor.

It seems that the problem stems from:

(https://github.com/ProteinsWebTeam/interpro7-client/blob/bd9f1dc619ef777fd54c5983f8aa28832320a4b8/src/components/ProteinViewer/style.css#L148)

Do we want it to be capped at 200px?

gustavo-salazar:
Mhh I think the cap to `400px` was a way to contain the problem. But the real issue was that a previous version of the nightingale track wasn't respecting the initial `width` parameter. I solved that bug and now `width` should work.

I'm adding a commit to fix everything to `300px` it's fixed so not responsive, but at least doesn't have that weird effect

gustavo-salazar:
Damn, I pushed that commit to the wrong branch: `taxonomy_tree_improvement`

gustavo-salazar:
OK I cherry pick the commit and ist's now here

matthiasblum (approved):
Looks good to me.

PR #651, opened 2024-08-02
TS migration of Help components

apolignano:
- Migrated the remaining JS components under the src/components/Help folder;
- Will need more time/guidance for the migration to the Visual framework.

See [IBU-11236](https://www.ebi.ac.uk/panda/jira/browse/IBU-11236)


apolignano:
Tests on GA failed multiple times even if local tests passed. It seems that it was being caused by lint-staged at commit time, splitting the (long) line below the // $FlowFixMe comment (changes were then pushed). Splitting the line manually beforehand and placing the // $FlowFixMe comment above the resulting second sub-line (vs. the first), solved the issue.

src/components/Help/TooltipAndRTDLink/index.tsx
```diff
@@ -1,22 +1,27 @@
...
 import config from 'config';
 import Link from 'components/generic/Link';
+
 // $FlowFixMe
 import Tooltip from 'components/SimpleCommonComponents/Tooltip';
 
-import { foundationPartial } from 'styles/foundation';
 import interproTheme from 'styles/theme-interpro.css'; /* needed for custom button color*/
 import fonts from 'EBI-Icon-fonts/fonts.css';
 import ipro from 'styles/interpro-new.css';
```
gustavo-salazar, line 11: As part of the migration I'm trying to remove all imports to `'styles/theme-interpro.css'` and `'styles/interpro-new.css'`to eventually be able to get rid of them. Moving any necesary styling in those into to local files, to try to have the styles closer to the code needed. And for things that are general styling moving it to general but specific files, e.g. `'styles/summary.css'` or `'styles/blocks.css'`

src/pages/Help/index.js
```diff
@@ -30,11 +32,15 @@ const Game = () => (
   <ContentFromRTD page="protein_families_game.rst" format="faq" />
 );
 
+// FlowFix me comment needed for async components
 const Documentation = loadable({
   loader: () =>
     import(
-      /* webpackChunkName: "help-documentation" */ 'components/Help/Documentation'
+      /* webpackChunkName: "help-documentation" */
```
gustavo-salazar, line 39: Not sure if webpack will be able to identify the chunk name if the line is split, rather use a `// prettier-ignore` for the block

src/components/Help/HelpBanner/index.tsx
```diff
@@ -62,7 +62,7 @@ export const HelpBanner = ({ topic }: Props) => {
   return (
     <Card
       title={
-        <span>
+        <span className={css('vf-card')}>
```
gustavo-salazar, line 65: I would not use `vf-card` here. I'm not really a fan of the drop shadow on this title. it gives the impression of a card inside a card.

src/components/Help/Documentation/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+import React from 'react';
+
+import config from 'config';
+import Link from 'components/generic/Link';
+import loadable from 'higherOrder/loadable';
+import { schemaProcessDataPageSection } from 'schema_org/processors';
+
+import fonts from 'EBI-Icon-fonts/fonts.css';
+import ipro from 'styles/interpro-new.css';
+import ebiGlobalStyles from 'ebi-framework/css/ebi-global.css';
```
gustavo-salazar, line 11: Similar to the comment in TooltipAndRTDLink: I'm trying to remove the use of `ebi-global.css`, most of the styling there overwrites foundation styles, so when foundation is removed it is not necessary. But we need to check it looks ok when it's out

src/components/Help/Documentation/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+    <section>
+      <SchemaOrgData
+        data={{
+          name: 'InterPro Documentation Page',
+          description:
+            'Includes links to the documentations for the different parts of our service',
+        }}
+        processData={schemaProcessDataPageSection}
+      />
+      <div className={f('row')}>
```
gustavo-salazar, line 33: `row` and `columns` are foundation styles I'm trying to move away, try to use `vf-stack` instead

gustavo-salazar (changes requested):
I have requested some changes in comments below. Mostly stuff that I forgot to mention about the migration of the styling.

apolignano:
- Removed global CSS from Documentation and RTDLink, added local styles;
- Added // prettier-ignore comment to avoid lint-staged pre-commit changes to lines below the FlowFixMe comment. Tests passing locally and on GH.

src/components/Help/Documentation/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+                // <li>
+                //   <Link href="//www.ebi.ac.uk/support/interpro-general-query">
+                //     <span
+                //       className={f('icon', 'icon-common')}
+                //       data-icon="&#x27a;"
+                //     />{' '}
+                //     Chat with an assistant{' '}
+                //   </Link>
+                // </li>
+              }
```
matthiasblum, line 195: What is that? :raised_eyebrow:

src/components/Help/Documentation/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+                    Citing Pfam
+                  </Link>
+                </li>
+              </ul>
+            </div>
+          </div>
+        </div>
+        <div className={f('vf-grid__col--span-1')}>
+          <div className={f('box-add', 'margin-top-xxlarge')}>
+            <h3 className={f('light')}>Contact us/Additional help</h3>
```
matthiasblum, line 183:
```suggestion
            <h3 className={f('light')}>Contact us</h3>
```

Text is too long for the sidebar.

src/components/Help/TooltipAndRTDLink/index.tsx
```diff
@@ -1,31 +1,35 @@
...
 import Tooltip from 'components/SimpleCommonComponents/Tooltip';
 
-import { foundationPartial } from 'styles/foundation';
-import interproTheme from 'styles/theme-interpro.css'; /* needed for custom button color*/
 import fonts from 'EBI-Icon-fonts/fonts.css';
-import ipro from 'styles/interpro-new.css';
+import local from './style.css';
+
+import cssBinder from 'styles/cssBinder';
+const f = cssBinder(local, fonts);
```
matthiasblum, line 13: It's not super important, but traditionally the `f` constant (which comes from Foundation, the old CSS framework) is renamed to `css`.

src/components/Help/TooltipAndRTDLink/style.css
```diff
@@ -0,0 +1,3 @@
+.rtd-link {
+  color: white !important;
+}
```
matthiasblum, line 3:
```suggestion
.rtd-link,  .rtd-link:active, .rtd-link:focus, .rtd-link:hover {
  color: white;
}
```

Try to avoid using `!important`, unless absolutely necessary. Also, the link needs to be white when active, hovered, and focused.

matthiasblum (commented):
Looks rather good to me.

I've added a few minor suggestions. Also, one file has not been migrated to TS (`src/pages/Help/index.js`). Why?

src/components/Help/Documentation/index.tsx
apolignano: Not sure, it was in the .js file as well.

apolignano:
[IBU-11236](https://www.ebi.ac.uk/panda/jira/browse/IBU-11236) is specifically about the components under src/components/Help. That is the Help page. I'll look into that.

src/components/Help/Documentation/index.tsx
matthiasblum: You can remove it. We do not plan to offer live support, nor we plan to create an AI assistant.

matthiasblum (approved):
Looks good to me.

src/components/Help/TooltipAndRTDLink/index.tsx
gustavo-salazar: I also do it because I don't know what extension. highlights the usage of `f` as a potential using a system function.

PR #656, opened 2024-08-19
AI Filter for entries

apolignano:
This PR adds an AI filter to the Browse page for InterPro and MemberDBs entries.

- The filter follows the same logic of the GOTerm filter.
- It's not shown if there's no LLM generated entries for a MemberDB (e.g. Pfam)

### DIFF
--- a/src/components/Entry/EntryListFilters/AIGeneratedFilter/index.tsx
+++ b/src/components/Entry/EntryListFilters/AIGeneratedFilter/index.tsx
@@ -20,9 +20,9 @@ import stylego from 'pages/style.css';
 const css = cssBinder(stylego, style);
 
 const categories = {
-  'Manually Curated': 'MC',
-  'AI-Generated and Reviewed': 'AI-R',
-  'AI-Generated and Unreviewed': 'AI-U',
+  Curated: 'curated',
+  'AI-Generated (reviewed)': 'ai-reviewed',
+  'AI-Generated (unreviewed)': 'ai-unreviewed',
 };
 type Categories = keyof typeof categories;
 
@@ -48,8 +48,8 @@ const AIGeneratedFilter = ({
 
   const _handleSelection = ({ target }: FormEvent) => {
     const value = (target as HTMLInputElement).value;
-    const { page, ai_category: _, cursor: __, ...restOfsearch } = search;
-    if (value !== 'All') restOfsearch.ai_category = value;
+    const { page, curation_status: _, cursor: __, ...restOfsearch } = search;
+    if (value !== 'All') restOfsearch.curation_status = value;
     goToCustomLocation?.({ ...customLocation, search: restOfsearch });
   };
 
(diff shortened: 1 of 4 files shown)


src/components/Entry/EntryListFilters/index.tsx
```diff
@@ -8,20 +8,26 @@ import EntryTypeFilter from './EntryTypeFilter';
...
+    <FiltersPanel>
+      <EntryTypeFilter
+        label={`${
+          mainDB === 'InterPro' ? 'InterPro' : 'Member Database Entry'
+        } Type`}
+      />
+      {mainDB !== 'InterPro' && <IntegratedFilter label="InterPro State" />}
+      {mainDB === 'InterPro' && <GOTermsFilter label="GO Terms" />}
+      {mainDB === 'InterPro' && <LatestFilter label="New entries" />}
+      <AIGeneratedFilter label={AILabel} changeLabel={setAILabel} />
```
matthiasblum, line 27: Unfortunately, changing the label to an empty string only hide the list of options, but there are still an instance of `FilterPanel` being created resulting in an empty space at the bottom:

image

The simplest is too always display the filter, regardless of the whether there are AI signatures or not.

src/components/Entry/EntryListFilters/AIGeneratedFilter/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+  if (!loading) {
+    terms.unshift(['All', NaN]);
+  }
+
+  const [showFilter, setShowFilter] = useState(true);
+
+  useEffect(() => {
+    if (terms.length > 0) {
+      // If loaded AI-generated reviewed and unreviewed counts will always be in position 2 and 3 of the terms obj
+      if (terms[2][1] == 0 && terms[3][1] == 0) {
```
matthiasblum, line 69: Are you sure about that?
`terms[0]` is always `['All', NaN]`, but other items are sorted by number of entries/signatures in the category. It's true that _at the moment_ there are always more curated entries than AI-generated, but you should not assume that it always be the case. 

You should check using the term, e.g.

```ts
if (terms.some(([term, count]) => (term === "AI-Generated (reviewed)" || term === "AI-Generated (unreviewed)") && count > 0)) {
  setShowFilter(true);
} else {
  setShowFilter(false);
}
```

src/components/Entry/EntryListFilters/AIGeneratedFilter/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+      } else {
+        setShowFilter(true);
+        changeLabel('AI-Generated Entries');
+      }
+    }
+  });
+
+  return (
+    showFilter && (
+      <div className={css('list-go', { stale: isStale })}>
```
matthiasblum, line 81:
```suggestion
      <div className={css('list-curation-status', { stale: isStale })}>
```

matthiasblum (changes requested):
Please see individual suggestions/comments.

src/components/Entry/EntryListFilters/index.tsx
apolignano: Yeah, I noticed. Unfortunately the way that page is structured seemed to not allow conditional rendering of the filter. Will look better into it to see if there's a better solution.

src/components/Entry/EntryListFilters/AIGeneratedFilter/index.tsx
apolignano: Missed the sorting part. Makes sense now. I'll fix this.

apolignano:
Now employing same logic of the other filters:

- If a filter count is 0, don't show it.
- Show count next to the "All" term


image

matthiasblum (approved):
LGTM. Feel free to merge.

PR #657, opened 2024-08-19
Reorganize tracks

apolignano:
This PR:

- Reorganizes tracks on the Protein summary page and hides (but makes expandable) all the ones considered "secondary". See [IBU-11247](https://www.ebi.ac.uk/panda/jira/browse/IBU-11247) for more;
- Adds representative family track;
- Adds disordered regions track;
- Changes the residue track so that domains are not shown;
- The logic for fetching PTM data is added but currently not used given the non-completeness of the PTM data;

### DIFF
--- a/src/components/ProteinViewer/grid.css
+++ b/src/components/ProteinViewer/grid.css
@@ -11,9 +11,10 @@
 .protvista-grid .view-options-wrap {
   margin-top: 30px;
   display: flex;
-  align-items: baseline;
+  align-items: center;
   justify-content: space-between;
   padding-bottom: 0.2rem;
+  margin-bottom: 10px;
 }
 .protvista-grid > header {
   background: #fafafa;
--- a/src/components/ProteinViewer/grid.css
+++ b/src/components/ProteinViewer/grid.css
@@ -12,10 +12,11 @@
   margin-top: 30px;
   display: flex;
   align-items: center;
-  justify-content: space-between;
   padding-bottom: 0.2rem;
   margin-bottom: 10px;
+  gap: 5px;
 }
+
 .protvista-grid > header {
   background: #fafafa;
   padding: 0.5rem 0.3rem;
--- a/src/components/ProteinViewer/grid.css
+++ b/src/components/ProteinViewer/grid.css
@@ -9,6 +9,7 @@
 }
 
 .protvista-grid .view-options-wrap {
+  margin-top: 30px;
   display: flex;
   align-items: baseline;
   justify-content: space-between;
--- a/src/components/ProteinViewer/index.tsx
+++ b/src/components/ProteinViewer/index.tsx
@@ -34,7 +34,6 @@ import tooltip from 'components/SimpleCommonComponents/Tooltip/style.css';
 import fonts from 'EBI-Icon-fonts/fonts.css';
 import RepresentativeDomainsTrack from './RepresentativeDomainsTrack';
 import ShowMoreTracks from './ShowMoreTracks';
-import FeatureShape from '@nightingale-elements/nightingale-track/dist/FeatureShape';
 
 TracksInCategory.displayName = 'TracksInCategory';
 Header.displayName = 'TracksHeader';
@@ -130,10 +129,10 @@ export const ProteinViewer = ({
 
   // List of "main" tracks to be displayed, the rest are hidden by default
   const mainTracks = [
-    'AlphaFold confidence',
+    'alphafold confidence',
     'representative domains',
     'representative families',
-    'variants',
+    'pathogenic variants',
     'disordered regions',
     'residues',
   ];
@@ -154,7 +153,6 @@ export const ProteinViewer = ({
     features: true,
     predictions: true,
     'match conservation': true,
-    'Clinical significance: pathogenic and likely pathogenic variants': true,
   });
 
   const categoryRefs = useRef<ExpandedHandle[]>([]);
@@ -266,8 +264,6 @@ export const ProteinViewer = ({
               >
                 {children}
               </Options>
-            </div>
-            <div className={css('track-sized')}>
               <ShowMoreTracks
                 showMore={showMore}
                 showMoreChanged={setShowMore}
@@ -290,7 +286,6 @@ export const ProteinViewer = ({
               []
               {(data as unknown as ProteinViewerData<ExtendedFeature>)
                 .filter(([_, tracks]) => tracks && tracks.length)
-
                 .map(([type, entries, component]) => {
                   entries.forEach((entry: ExtendedFeature) => {
                     entry.protein = protein.accession;
@@ -305,15 +300,19 @@ export const ProteinViewer = ({
                   }
 
                   if (type == 'residues') {
-                    let residuesEntries: ExtendedFeature[] = [];
-                    residuesEntries = [...entries];
-                    for (let i = 0; i < residuesEntries.length; i++) {
-                      delete residuesEntries[i].entry_protein_locations;
-                      residuesEntries[i].type = 'residue';
-                      residuesEntries[i].locations = residuesToLocations(
-                        residuesEntries[i].residues,
-                      );
-                    }
+                    const residuesEntries: ExtendedFeature[] = [];
+                    entries.map((entry) => {
+                      const tempFeature: ExtendedFeature = {
+                        accession: entry.accession,
+                        name: entry.name,
+                        protein: entry.protein,
+                        source_database: entry.source_database,
+                        type: 'residue',
+                        locations: residuesToLocations(entry.residues),
+                      };
+                      residuesEntries.push(tempFeature);
+                    });
+
                     entries = residuesEntries;
                   }
 
--- a/src/components/ProteinViewer/index.tsx
+++ b/src/components/ProteinViewer/index.tsx
@@ -139,25 +139,25 @@ export const ProteinViewer = ({
   ];
 
   const [hideCategory, setHideCategory] = useState<CategoryVisibility>({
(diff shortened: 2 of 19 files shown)


src/components/ProteinViewer/index.tsx
```diff
@@ -122,10 +124,39 @@ export const ProteinViewer = ({
...
+    'binding site': true,
+    ptm: true,
+    unintegrated: true,
+    'other features': true,
     'other residues': true,
-    'external sources': true,
+    features: true,
+    predictions: true,
+    'match conservation': true,
+    'Clinical significance: pathogenic and likely pathogenic variants': true,
```
matthiasblum, line 157:
```suggestion
    'pathogenic and likely pathogenic variants': true,
```

I don't think we need the prefix the track with Clinical significance.

src/components/ProteinViewer/utils.ts
```diff
@@ -3,22 +3,25 @@ import { toPlural } from 'utils/pages/toPlural';
...
-  const flatDomains = [];
-  for (const domain of domains) {
+  const flatRepresentativeData = [];
+
+  for (const entry of entries) {
     const { accession, short_name, name, source_database, integrated, chain } =
-      domain;
-    if (domain[locationKey] === null) {
+      entry;
+    if (entry[locationKey] === null || entry.type != type) {
```
matthiasblum, line 17:
```suggestion
    if (entry[locationKey] === null || entry.type !== type) {
```

src/components/Related/DomainsOnProtein/DomainsOnProteinLoaded/index.tsx
```diff
@@ -13,11 +13,19 @@ const ProteinViewer = loadable({
   loading: null,
 });
 
+// 0A017SEX7 is a good example
 const UNDERSCORE = /_/g;
 const FIRST_IN_ORDER = [
   'representative_domains',
-  'secondary_structure',
+  'representative_families', // coming soon
```
matthiasblum, line 20:
```suggestion
  'representative_families',
```

src/components/Related/DomainsOnProtein/DomainsOnProteinLoaded/index.tsx
```diff
@@ -13,11 +13,19 @@ const ProteinViewer = loadable({
...
+  'variants',
+  'disordered_regions', // data coming from (?)
```
matthiasblum, line 22:
```suggestion
  'intrinsically_disordered_regions',
```

It's coming for the protein endpoint with the `extra_features` modifier, e.g. `/api/protein/UniProt/P04637/?extra_features`.

src/components/Related/DomainsOnProtein/DomainsOnProteinLoaded/index.tsx
```diff
@@ -13,11 +13,19 @@ const ProteinViewer = loadable({
...
+  'residues',
+];
+
+const LASTS_IN_ORDER = [
   'family',
+  'secondary_structure',
```
matthiasblum, line 28: `LASTS_IN_ORDER` is to force some tracks to be at the end. Tracks that are neither in `FIRST_IN_ORDER` nor in `LASTS_IN_ORDER` will be in the middle, which is I think the desired behaviour.

src/components/Related/DomainsOnProtein/DomainsOnProteinLoaded/index.tsx
```diff
@@ -26,14 +34,13 @@ const FIRST_IN_ORDER = [
   'binding_site',
   'ptm',
   'unintegrated',
-];
-const LASTS_IN_ORDER = [
+  'other_features',
```
matthiasblum, line 37: See previous comment.

src/components/Related/DomainsOnProtein/DomainsOnProteinLoaded/index.tsx
```diff
@@ -43,8 +50,8 @@ export const byEntryType = (
     if (b.toLowerCase() === label) return 1;
   }
   for (const l of LASTS_IN_ORDER) {
-    if (a.toLowerCase() === l) return 1;
-    if (b.toLowerCase() === l) return -1;
+    if (a.toLowerCase() === l) return -1;
+    if (b.toLowerCase() === l) return 1;
```
matthiasblum, line 54: I am puzzled. Why does the `-1` and `1` had to be switched?

src/components/Related/DomainsOnProtein/index.tsx
```diff
@@ -155,8 +170,25 @@ const DomainOnProteinWithoutData = ({
...
+      (k) =>
+        (k as MinimalFeature).accession ==
+        'Mobidblt-Consensus Disorder Prediction',
+    );
+    return mobiDBLiteEntries;
+  };
+
   if (dataFeatures && !dataFeatures.loading && dataFeatures.payload) {
     mergeExtraFeatures(mergedData, dataFeatures?.payload);
+    mergedData['disordered_regions'] = filterMobiDBLiteFeatures(
```
matthiasblum, line 189:
```suggestion
    mergedData['intrinsically_disordered_regions'] = filterMobiDBLiteFeatures(
```

src/components/Related/DomainsOnProtein/DomainsOnProteinLoaded/index.tsx
```diff
@@ -69,6 +80,10 @@ export const makeTracks = ({
   if (other) mergedData.other_features = other;
   if (representativeDomains?.length)
     mergedData.representative_domains = representativeDomains;
+  if (representativeFamilies?.length)
+    mergedData.representative_families = representativeFamilies;
+  if (disorderedRegions?.length)
+    mergedData.disorderedRegions = disorderedRegions;
```
matthiasblum, line 86:
```suggestion
    mergedData.intrinsically_disordered_regions = disorderedRegions;
```

src/components/ProteinViewer/index.tsx
```diff
@@ -122,10 +124,39 @@ export const ProteinViewer = ({
...
+  // State variable to show/hide "secondary" tracks
+  const [showMore, setShowMore] = useState(false);
+
+  // List of "main" tracks to be displayed, the rest are hidden by default
+  const mainTracks = [
+    'AlphaFold confidence',
+    'representative domains',
+    'representative families',
+    'variants',
+    'disordered regions',
```
matthiasblum, line 137:
```suggestion
    'intrinsically disordered regions',
```

matthiasblum (changes requested):
Thanks! This looks good overall.

I've made a few suggestions. 

Besides them, there are two required changes.

**Residues**

Unfortunately, stacking all the modified residues in the tooltip can lead to this kind of monstrosity:

image

... which is something that is already an issue in the live InterPro website. We could display more than one residue at a time in the tooltip, but it should be something that is more user friendly, e.g. a list of the individual positions (`269, 496, 498, etc.`). 

Also, the labels on the right of the track don't seem the be links.

**Intrinsically disordered regions**

If you look at [P05067](https://www.ebi.ac.uk/interpro/protein/UniProt/P05067/), you'll notice that there are three MobiDB-lite hits:
- Consensus Disorder Prediction
- Polar
- Polyampholyte

The two last tracks are additional annotations which are subregions of the first one. If possible, it would great to display them nested under the first one, like tracks in the "Family" or "Domain" category (one parent track containing one or more sub tracks that can be hidden when clicking on the parent track).

matthiasblum (changes requested):
Thanks.

As we're slowly transitioning to highlight matches from individual member database signatures over InterPro entries, you also need to refactor how matches are displayed in the, "Domain", "Family", "Homologous Superfamily", and "Conserved Site" categories.

The idea is to stop showing the InterPro matches, with the individual signature matches nested within. Instead, we only display the signatures matches (with a link to the entry in the tooltip if they are integrated).

**How to know in which category (Domain, Family, etc.) the match should be included?**
Use the value of `type` for the signature match, instead the signature is integrated in an InterPro entry, in which case the `type` of the InterPro entry should be used.

**How to organise matches in each category**
From left to right: order matches by their leftmost and rightmost boundaries, _but_ group matches integrated in the same InterPro entry together (one per line, like in the current website).

**What about the "Unintegrated" category?**
Ditch it. Its regions will now be displayed based in the `type` of the individual matches.

**What about "Other features"?**
Oh boy.
- Matches from coils, TMHMM, Phobius, and SignalP: new category labelled "Coiled-coils, signal peptides and transmembrane regions"
- Pfam-N: new "Pfam-N" category (for now, we might move them closer to Pfam matches later)
- AntiFam: new "Spurious protein" category, displayed by default (like MobiDB matches) :warning: 
- ELM: new "Short linear motifs" category
- FunFam: don't touch them for now. Need to think where is the best place for these ones

When this is done, we' ll also take care of the "External sources".

Good luck ;)

src/components/ProteinViewer/grid.css
```diff
@@ -9,10 +9,12 @@
 }
 
 .protvista-grid .view-options-wrap {
+  margin-top: 30px;
   display: flex;
-  align-items: baseline;
+  align-items: center;
   justify-content: space-between;
```
matthiasblum, line 15: Remove this rule so we can push the `Options` component to the right.

```suggestion
```

Then add the following rule under `.view-options { ... }` in `src/components/ProteinViewer/Options/styles.css`:

```css
  margin-left: 0 0.5rem 0 auto;
```

matthiasblum (approved):
I've pushed two minor commits. 

Looks good to me. Feel free to merge. Thanks!

PR #660, opened 2024-08-22
PTMs track

apolignano:
Branched off of newer version of reorganize-tracks. Added logic for PTM data fetching and processing, related track, tooltips and labels.


matthiasblum (changes requested):
So the data looks OK, but there are a few changes needed before it can be merged:

- resolve conflicts
- switch from `<proteinsAPI>/proteomics-ptm/<accession>` to `<proteinsAPI>/proteomics/ptm/<accession>`
- group PTMs by name (e.g. Phosphorylation) so the name can be used as label on the right. Currently it looks that all PTM are already displayed on the same track, but if you have two PTMs with different names, I am not sure what would happen with the current code. What would we need is to have two tracks, one for each name.
- the tooltip should include the start/end positions. The current tooltip contains

> Peptide: TGQAPGY[S]TAANK (starts at 41)
> Phosphorylation on S (8)
> Source: PRIDE

I think it should be:

> Peptide: TGQAPGYSYTAANK (41-54)
> Phosphorylation on S (48)
> Source: PRIDE

(The `S` in the peptide sequence should keep being in bold)

matthiasblum (changes requested):
Almost there.

I spotted a bug when hovering the match of a PTM _entry_ for `A0A096LNW5`, probably due to a clash between the data/component for the PTM entry and the PTM data retrieved from the Proteins API.

apolignano:
- Fixed bug that prevented PTMs from InterPro entries to be displayed correctly. Now showing the standard entry popup instead of the PTM one; 
- Introduced some logic to have PTMs coming from the API and from InterPro to be shown under the same section (testing needed, either with dummy data or with protein that has both);

matthiasblum (changes requested):
The bug that happened when browsing a protein with a PTM InterPro entry (e.g. UniProt accession `A0A096LNW5`) seems fixed, but I still cannot see the signature match (InterPro entry + member databse signature) _and_ the PTM (from the Proteins API) when browsing proteins having both, e.g. `Q9Y2Q0`.

Also, please use "Post-translational modifications" instead of "Ptm" for the category/section label.

matthiasblum (approved):
All good. Feel free to merge. Thank you!

PR #668, opened 2024-10-15
Curation and Signature subpages

apolignano:
This PR removes the Curation section and moves its contents as follows:
- curation information in Overview section, top table;
- HMM information in ex Signature section, now renamed HMM Profiles;

image
image


src/components/Entry/Summary/MemberDBSubtitle/index.tsx
matthiasblum, line 1: A message could be added so users visiting the `curation` subpage are informed that the information has been moved to `Overview` and `Profile HMM`.
Deleted pages are defined in `src/utils/removed-pages/index.tsx`

matthiasblum (changes requested):
A few minor things to fix. Thanks!

src/components/Entry/HmmModels/index.js
```diff
@@ -77,6 +94,52 @@ const HmmModelSection = ({ logo } /*: {logo: {}} */) => {
...
+                          data-icon="&#xf019;"
+                        />{' '}
+                        Download
+                      </Link>{' '}
+                      the raw HMM for this family
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+            </>
```
matthiasblum, line 139: I only now realize that the HMM download link won't be displayed if the details are not available.

Would it be possible to always show the Download row (with the link, as the first row of the table, and to show the two other rows (HMM commands and gathering thresholds) if they are available?

matthiasblum (changes requested):
Almost there. Just a one or two things to address. Also, the `src/subPages/Curation` directly could be deleted.

src/components/Entry/HmmModels/index.js
apolignano: Sure. The download URL also needs to be changed. It seems that it's always pointing to pfam. Changing that to

```jsx
<Link href={`${config.root.API.href}/entry/${payload?.metadata?.source_database}/${payload?.metadata?.accession}?annotation=hmm`} download={`${payload?.metadata?.accession || 'download'}.hmm.gz`}>
```

matthiasblum (approved):
LGTM. I just pushed a commit that fixes the external link to http://www.sequenceontology.org.

Feel free to merge when tests complete. Thanks!

PR #674, opened 2024-11-04
Isoform and clan fixes

apolignano:


src/components/Protein/Isoforms/Viewer/index.tsx
```diff
@@ -70,18 +70,45 @@ const features2protvista = (features: FeatureMap) => {
...
+    'locations',
+    'family',
+  );
+
+  const homologous_superfamily = sortedCategories.filter(
+    (entry) => entry[0] == 'homologous superfamily',
+  )[0];
+
+  sortedCategories.map((entry) => {
+    if (entry[0] === 'domain') {
```
matthiasblum, line 91: If this is the type (as returned by the API and found in `metadata.type`), then the type _repeat_ should be added (representative domains are for signatures of type domain _and_ repeat).

matthiasblum (commented):
Potentially one thing that needs to be changed.

Rest is good.

matthiasblum (approved):

PR #678, opened 2024-11-12
ShowMore in redux state

apolignano:


apolignano:
- The ShowMore button state is now saved in the Global Redux State, allowing users to keep the preferred view.
- The styling of the toggle button could be up for discussion to make it as similar as possible to the other VF components.

matthiasblum (changes requested):
The switch works as intended, but there are a few things that could be improved

1. it looks like you have re-implemented a switch button. Can you use the [`ToggleSwitch`](https://github.com/ProteinsWebTeam/interpro7-client/blob/master/src/components/ToggleSwitch/index.tsx) component instead?
2. The More/Less option isn't so clear. We could add a label _Feature display_ in front of the switch and the options of the switch would be _Overview_ and _Full_.

apolignano:
Unfortunately, the ToggleSwitch component doesn't look great when positioned there, as it only accepts three sizes and none of them match the Option menu button. I've tried to pass an height prop but you end up having to play around too much with margins, CSS vars, etc.. Tried the label too, and it looks even worse IMO. 

Less and More are renamed to Overview and Full and the color-changing behavior of ToggleSwitch has been implemented.

Screenshot 2024-11-22 at 12 26 40

Screenshot 2024-11-22 at 12 28 09

Let me know if it works!

src/components/ProteinViewer/ShowMoreTracks/index.tsx
```diff
@@ -20,44 +19,47 @@ type Props = {
...
-      </button>
+        <div
+          className={css('showmore-toggle-inner', showMore ? 'moved' : '')}
+        />
+        <span style={{ marginLeft: showMore ? '7px' : '25px' }}>
+          {showMore ? 'Full View' : 'Overview'}
+        </span>
+      </div>
     </div>
   );
```
matthiasblum, line 64: Minor style changes, and add a label in front of the switch (not the best UI, but I think we need a label):

```ts
  return (
    <>
      Feature display:
      <div onClick={handleClick}>
        <div
          className={css('showmore-toggle-outer', showMore ? 'full' : 'overview')}
        >
          <div
            className={css('showmore-toggle-inner', showMore ? 'moved' : '')}
          />
          <span style={{ marginLeft: showMore ? '5px' : '33px' }}>
          {showMore ? 'Full' : 'Overview'}
        </span>
        </div>
      </div>
    </>
  );
```

src/components/ProteinViewer/ShowMoreTracks/style.css
```diff
@@ -1,3 +1,39 @@
 .showmore-btn {
   margin: 0;
 }
+
+.showmore-toggle-outer {
+  display: flex;
+  align-items: center;
+  width: 105px;
```
matthiasblum, line 8:
```suggestion
  width: 118px;
```

src/components/ProteinViewer/ShowMoreTracks/style.css
```diff
@@ -1,3 +1,39 @@
...
+
+.showmore-toggle-outer.full {
+  background-color: #1779ba;
+}
+
+.showmore-toggle-inner {
+  position: absolute;
+  top: 5px;
+  left: 5px;
+  width: 15px;
```
matthiasblum, line 31:
```suggestion
  width: 28px;
```

matthiasblum (commented):
Looks better. I suggested a few changes that I think could improve the UI. Please have a look and let me know if you agree with them.

PR #682, opened 2024-12-03
AlphaFold casing

apolignano:


src/components/BreadCrumbs/index.js
```diff
@@ -201,7 +201,13 @@ const BreadCrumbForEntityDetail = (
     filters = [],
   } /*: {location: Location, detail?: string, filters: Array<string>} */,
 ) => {
-  const detailToRender = !filters.length && !detail ? ' Overview' : detail;
+  let detailToRender = !filters.length && !detail ? ' Overview' : detail;
```
matthiasblum, line 204: A bit clearer/intuitive (positive conditions first, avoids the potential confusion of double negations).

```suggestion
  let detailToRender = filters.length > 0 || detail ? detail : 'Overview';
```

matthiasblum (approved):
LGTM. I just made a small suggestion for a line that was there even before this PR.

Quick question: would it be possible to migrate the `BreadCrumbs` component to tsx? If yes, please do, otherwise feel free to merge.

matthiasblum (changes requested):
Tests fail after the TSX migration, because 

```
// $FlowFixMe
```

should be placed before this line:

https://github.com/ProteinsWebTeam/interpro7-client/blob/23d6819c3de3e642a73b578d6958b3ca54550682/src/pages/index.js#L12

PR #695, opened 2025-02-11
Hover bug in Alphafold Protein Viewer

apolignano:


apolignano:
The bug, spotted on interpro/protein/reviewed/Q2FE14/alphafold/, cause the "Families" and "Domains" track to re-collapse upon hovering on one of the matches. That seemed to happen because of the state update due to the selection on the 3D model structure based on the match. We now avoid recomputing visibilityCategories/showMore flags for that viewer.

src/components/ProteinViewer/index.tsx
```diff
@@ -372,19 +379,21 @@ export const ProteinViewer = ({
...
+                  showMore={showMore}
+                  changeSettingsRaw={changeSettingsRaw}
+                  showMoreChanged={setShowMore}
+                  setHideCategory={setHideCategory}
+                  switchCategoryVisibilityShowMore={
+                    switchCategoryVisibilityShowMore
+                  }
+                  hideCategory={hideCategory}
+                />
+              )}
```
matthiasblum, line 396: This causes the switch to be displayed on structure pages (several times, if there are several chains), e.g. `/interpro/structure/PDB/1t2v/`. On this page, I experienced the same bug as on AlphaFold pages (section collapsing on hover), although it seems to occur more randomly so it might be difficult to reproduce.

Also, on the structure pages, switching between the overview and full mode throws several TypeErrors due to `event.detail` being undefined at these lines:

https://github.com/ProteinsWebTeam/interpro7-client/blob/be3b06606dd3ebb14112dd890e7b9bc996b04ebc/src/components/Structure/ViewerAndEntries/index.tsx#L105

https://github.com/ProteinsWebTeam/interpro7-client/blob/be3b06606dd3ebb14112dd890e7b9bc996b04ebc/src/components/Related/DomainEntriesOnStructure/ProteinViewerLoaded/index.tsx#L26

matthiasblum (changes requested):
Thanks. Seems to correctly address the bug on AlphaFold pages, but as a negative impact on the structure pages (see comment below).

apolignano:
I introduced the viewerType prop in the other viewer. This way, both types of Protein Viewers (structure and AlphaFold) will have the Display Mode section hidden and will just pick up the choice the user has already made on the ProteinViewer of the protein page or the InterPro Scan Search result page.

matthiasblum (approved):
LGTM. Feel free to merge, then please deploy. When deploying, please remember to:

- deploy on failover VM
- confirm that the bug doesn't occur on the failover VM
- switch off live API (traffic is rerouted to failover)
- deploy on live VM
- restart live API
- do a quick check

PR #701, opened 2025-02-19
Add link to page where download job was initiated

matthiasblum:
Currently, when listing [download jobs](https://www.ebi.ac.uk/interpro/result/download/#table), the first column is the URL to the API page used to fetch records. This can be confusing for non-technical users, but also can be misleading as results may span over several pages instead of all being returned by one page.

This PR adds some logic to keep track of the URL of the page where the download job was created.

Example to test:

1. Go to `/entry/InterPro/?type=active_site`
2. Open the _Download_ dropdown
3. Click _Generate TSV_
4. Go to `/result/download/` or `Results > Your downloads`

Live website:

image

This branch:

image

The first column is now a link to the _website_ page where the download job was created. The link to the first page of the API has been moved under "Actions".

### DIFF
--- a/src/actions/creators/index.ts
+++ b/src/actions/creators/index.ts
@@ -224,6 +224,7 @@ export const downloadURL = (
   fileType: DownloadFileTypes,
   subset: boolean,
   endpoint: string,
+  originURL?: string,
 ) =>
   ({
     type: types.DOWNLOAD_URL,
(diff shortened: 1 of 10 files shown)


apolignano:
Looks good. I was able to reproduce the example.

But, when the user already has saved jobs with the old format, the page returns an error because the renderer can't process "originURL", undefined in that case. Added a little piece of logic that transforms the localID to the originURL with a regex.

apolignano (approved):

PR #702, opened 2025-02-19
Structure highlight fixes

matthiasblum:
There are cases where highlighting a domain on a PDB structure doesn't work, e.g. with [`3pzd`](https://www.ebi.ac.uk/interpro/structure/PDB/3pzd/#table):

<details>
<summary>Screenshot without highlight</summary>

image
</details>

<details>
<summary>Screenshot with highlight (any domain)</summary>

image
</details>

This is due to the residues _on the 3D structure_ having different positions. In the case of `3pzd` ([link to PDBe](https://www.ebi.ac.uk/pdbe/entry/pdb/3pzd/protein/1)), the first four amino acids of the sequence are `EFDT` (Glu - Phe - Asp - Thr). You would expect the position of the `E` glutamate to be `1`, but...

image

its position is `1501`.

I've added an alternative mapping to `entry_structure_locations.fragments` objects: `auth_start` and `auth_end`, which stores the positions of a match on the 3D structure, for instance:

```json
[
    {
        "fragments": [
            {
                "start": 89,
                "end": 193,
                "dc-status": "CONTINUOUS",
                "auth_start": 1589,
                "auth_end": 1693
            }
        ],
        "representative": false,
        "model": "PF00784",
        "score": null
    }
]
```

This PR uses `auth_start` and `auth_end`, when available, to correctly highlight domains.

This PR also fixes the highlight of clicked domains on AlphaFold structures, and highlights the domain in the colour they have on the protein viewer, instead of colouring them in green.

### DIFF
--- a/src/components/Structure/ViewerAndEntries/ProteinViewerForAlphafold/index.tsx
+++ b/src/components/Structure/ViewerAndEntries/ProteinViewerForAlphafold/index.tsx
@@ -118,10 +118,14 @@ const ProteinViewerForAlphafold = ({
           default:
             break;
         }
-      } else {
-        if (eventType === 'mouseout') {
-          setHoverSelection([]);
-          isHovering.current = false;
+      } else if (eventType === 'mouseout') {
+        setHoverSelection([]);
+        isHovering.current = false;
+      } else if (eventType === 'click') {
+        if (fixedSelectionRef.current.length) {
+          setFixedSelection([]);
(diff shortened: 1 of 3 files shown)


apolignano:
Looks great. 
Also tested both issues with the highlight color (always green/disappearing after click) and they're indeed gone.
We can merge this.

apolignano (approved):

PR #703, opened 2025-02-20
Merge dev into master (unintegrated/empty section bug)

apolignano:


matthiasblum (changes requested):
Thanks. Matches from unintegrated are now displayed, but so are PIRSR matches, e.g.,

image

(Example sequence, display mode set to full).

Job: `/interpro/result/InterProScan/iprscan5-R20250220-120841-0406-19174131-np2/`

apolignano:
That is fixed in the branch with the new reorganization and refactoring. Will be fixed when we merge after that PR review (including InterPro-N).

matthiasblum (approved):
LGTM. Thanks for the quick fix! Feel free to merge and deploy.

PR #711, opened 2025-03-17
Integration of BFVD 

apolignano:
This PR adds BFVD support. 

- For proteins with BFVD entries:
  - we display a BFVD menu entry like we do with AlphaFold;
  - The molstar 3D structure viewer loads its .pdb file from the BFVD servers;
  - A plDDT score track is displayed.
  - The match highlighting functionality stays exacly the same;

- BFVD API Status was added to the right-side menu.


src/components/Structure3DModel/3DModel/index.tsx
```diff
@@ -76,15 +78,42 @@ const AlphaFoldModel = ({
...
+        if (res.status == 200) {
+          setIsPDBAvailable(true);
+        }
+      });
+    } else {
+      // If not using BFVD, assume AlphaFold models are available
+      setIsPDBAvailable(true);
+      setIsPDBLoading(false);
+    }
+    setIsPDBLoading(false);
```
matthiasblum, line 102: Call `setIsPDBLoading()` inside the callback of `fetch`, so after the promise has resolved, otherwise `setIsPDBLoading(false)` is called before the `fetch` returns anything, so we end up with `isPDBLoading` set to `false` and `isPDBAvailable` set to `true`, and the warning message (_Structure Viewer currently not available for this entry_) is temporary displayed.

```suggestion
        if (res.ok) {
          setIsPDBAvailable(true);
        }
        setIsPDBLoading(false);
      });
    } else {
      // If not using BFVD, assume AlphaFold models are available
      setIsPDBAvailable(true);
      setIsPDBLoading(false);
    }
```

src/components/Structure3DModel/3DModel/index.tsx
```diff
@@ -93,29 +122,67 @@ const AlphaFoldModel = ({
...
+    return <Loading />;
+  } else {
+    if (!isPDBAvailable) {
+      return (
+        <Callout type="warning">
+          Structure Viewer currently not available for this entry.
+        </Callout>
+      );
+    }
+  }
```
matthiasblum, line 136: Shouldn't this be wrapped in a `if (bfvd) { ... }` since it's specific to BFVD?

src/store/enhancer/status-middleware/index.js
```diff
@@ -24,16 +24,26 @@ const checkStatusesAndDispatch = async function (
   if (!browser) return;
   for (const endpoint of Object.keys(servers)) {
     const endpointSettings = settings[endpoint];
+
+    let testEndPoint = '';
+    switch (endpoint) {
+      case 'alphafold':
+        testEndPoint = 'api/prediction/Q5VSL9';
+      case 'bfvd':
+        testEndPoint = 'api/cluster/A0A2Z4HFS2';
```
matthiasblum, line 33:
```suggestion
        testEndPoint = 'api/prediction/Q5VSL9';
        break;
      case 'bfvd':
        testEndPoint = 'api/cluster/A0A2Z4HFS2';
        break;
```

matthiasblum (changes requested):
Thanks. A few things to address, see comments/suggestions.

Something I noticed is that on BFVD subpages, there is a call to AlphaFold that is systematically attempted, e.g. `/interpro/protein/unreviewed/A0A8S5TK01/bfvd/`. The opposite doesn't seem to be true, i.e. on AlphaFold subpages, no query to the BFVD API is done. Ideally, we wouldn't query the AlphaFold API if we know that the predicted structure comes from BFVD.

Also, the protein Overview page shows links to AlphaFold and Foldseek when an AlphaFold structure is available, e.g. https://www.ebi.ac.uk/interpro/protein/UniProt/A0A000/.

The same should be done for proteins with a BFVD structure.

src/components/EntryMenu/index.tsx
```diff
@@ -156,20 +157,22 @@ export class EntryMenuWithoutData extends PureComponent<LoadedProps, State> {
...
+              // activeClass={e.activeClass}
+              name={e.name}
+              data={data}
+              counter={e.counter}
+              isFirstLevel={!mainAccession}
+              usedOnTheSide={usedOnTheSide}
+              collapsed={this.state.isCollapsed}
+            />
+          );
+        })}
```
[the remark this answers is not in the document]
apolignano: Probably during testing I was logging infos and I needed that to be a function with an explicit return.

src/components/Structure3DModel/3DModel/index.tsx
apolignano: Yup, it was that way before the Loading fix commit and tried to fix in a different way. This suggestion doesn't work either because the `then` throws an error before executing  setIsPDBLoading(false). Handling it properly with a `catch` now.

matthiasblum (changes requested):
Things to change:

1. pLDDT track shorter than the sequence for `A0A2Z4HFS2`

image

2. External link associated to the `pLDDT` label should point to BFVD, not AlphaFold

3. If a BFVD prediction is available, links to BFVD and Foldseek should be displayed under "External Links" on the protein overview page, like we do for AlphaFold.

matthiasblum (approved):

PR #712, opened 2025-03-20
Hide expiry date if in the past

matthiasblum:
If you open an InterProScan search that you haven't started yourself, your browser doesn't have job metadata like the creation date/time, which causes the the expiry date to epoch + one week (so 8 January 1970).

Example: https://www.ebi.ac.uk/interpro/result/InterProScan/iprscan5-R20250320-100548-0158-19647556-p1m/#table

This small PR hides the expiry date if it's in the past.


apolignano (approved):

apolignano:
Tested. Looks good. Feel free to merge!

PR #718, opened 2025-04-11
InterPro-N callout

matthiasblum:
Noticed that the InterPro-N callout isn't displayed on the AlphaFold and BFVD subpages. This PR fixes it.


apolignano (approved):
Looks good. Let's merge.

PR #721, opened 2025-04-15
BFVD and Repeat fixes

apolignano:
This PR addresses the following fixes:
- Repeats were .. repeated.. because of a mismatch domain != repeat between child and parent element. (`A0A084B9Z8`); 
- BFVD fixed highlighting on click now works;
- Some matches aren't displayed when opening the page. Didn't experience the problem again. Will have to check if that happens once published on `wwwdev`.


matthiasblum (approved):
LGTM. Thanks for the quick fix!

Merging & deploying on `wwwdev` :rocket:

PR #722, opened 2025-04-22
Minor change to domain viewer margins

matthiasblum:
This PR changes how bottom margins are used for the domain viewer, i.e. only use larger margins, required to show the Options menu with the InterPro-N options, when we have non-structure proteins.

It also slightly reduces the domain viewer and its title.


apolignano (approved):
Looks good. Let's merge.

Issue #731, opened 2025-04-25
Representative domains in InterProScan results


blazaropinto:
I just run InterProScan with the default settings on the example sequence on the InterPro website, please see screenshot below (note PANTHER -always type family- seems to be within the representative domains)
(I tried both in Mozilla and Chrome)

![Image](https://github.com/user-attachments/assets/63dd9b72-9124-46d1-8612-fc75ea7d95dc)

PR #733, opened 2025-04-25
Fix IPScan representative data

apolignano:
Addresses #731


src/types/external-payloads.d.ts
```diff
@@ -7,6 +7,7 @@ type Iprscan5Signature = {
     version: string;
   };
   entry: Iprscan5Entry;
+  type: string;
```
matthiasblum, line 10:
```suggestion
  type?: string;
```

matthiasblum (approved):
Thanks. Minor suggestion to include, then please merge and open a PR for `dev` -> `master`.

PR #735, opened 2025-04-29
Improve Search Bar in Header

apolignano:
This PR addresses the following:
- incomplete/incorrect data loading when redirection to a protein page happens after pressing enter in the search box in the header; 
- no automatic redirection/search initiation while the user is typing. The search action needs to be confermed by either pressing enter or by clicking on the search button;
- disabled usage of "search" history hints that caused the previously typed string to always reappear in the search box;


apolignano:
- Parametrized Link component in Header to make behavior consistent between click/enter;
- Search page is not briefly shown anymore. Also removed shouldRedirect. Not needed.

matthiasblum (approved):
LGTM, thanks

PR #736, opened 2025-04-29
Improved options and unintegrated matches type for IPScan Jobs

apolignano:
This PR addresses the following:
- the unintegrated section of the IPScan Job results still contained some matches due to missing type in data coming in from IPScan. Now that the type is available, we can place them in the correct section.
- advanced options for job submission improvements, as outlined in: [IBU-11828](https://embl.atlassian.net/browse/IBU-11828)

### DIFF
--- a/src/components/IPScan/AdvancedOptions/index.tsx
+++ b/src/components/IPScan/AdvancedOptions/index.tsx
@@ -20,7 +20,7 @@ const mdb1Values = new Set([
   'CDD',
   'HAMAP',
   'Panther',
-  'PfamA',
+  'Pfam',
   'PIRSF',
   'PRINTS',
   'PrositeProfiles',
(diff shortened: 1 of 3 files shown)


matthiasblum (changes requested):
For the advanced options of InterProScan when submitting a job:

- what I meant in [IBU-11828](https://embl.atlassian.net/browse/IBU-11828) is that the `SignalP` alias should be removed, not the three individual SignalP applications (`SignalP_EUK`, `SignalP_GRAM_POSITIVE`, and `SignalP_GRAM_NEGATIVE`). These three should be left so users can decide which analyses to run based on their dataset.
- `Pfam` is now displayed under _Other sequence features_. It should still be under _Families, domains, sites & repeats_.
- Selecting the `Pfam` checkbox submits the job with `PfamA`, not `Pfam`

The use of `signature.type` in the InterProScan results seems to work well.

Thanks.

apolignano:
- Oops. Read the SignalP one wrong. Fixed now.
- Apparently IPScan doesn't allow "Pfam" as a value. Only "PfamA" is accepted.
I guess either we change it on IPScan or keep it that way on the client (?).
  ```
  <error>
   <description>Invalid parameters: 
  Applications -> Value for "appl" is not valid: Currently "Pfam" but should be one of the restricted values: NCBIfam, SFLD, Phobius, SignalP, SignalP_EUK, SignalP_GRAM_POSITIVE, SignalP_GRAM_NEGATIVE... (check parameter documentation for more details)</description>
  </error>
  ```

matthiasblum (approved):
For `PfamA` being submitted and `Pfam` not being accepted: when you run InterProScan 5 locally, both `Pfam` and `PfamA` are accepted (`PfamA` is a deprecated name and it automatically replaced by `Pfam`). I guess we never asked the Job Dispatcher service to add `Pfam`.

As we don't plan to maintain InterProScan 5 for much longer, it's better to keep using `PfamA` when submitting sequences.

matthiasblum (approved):
LGTM.

Please merge and open a PR `dev` -> `master`.

PR #743, opened 2025-05-19
Don't show highlighting menu if there are no entries

apolignano:


matthiasblum (changes requested):
Could you add a message that says

> No entry matches this structure

where the domain viewer would be, like we have a message showing "No entry matches this protein" for proteins without any matches?

matthiasblum (approved):
LGTM

PR #746, opened 2025-05-30
Include short_name column in TSV download

apolignano:
This PR addresses [IBU-11884](https://embl.atlassian.net/browse/IBU-11884).

Given the separate strategies that are used to get data for the table and the TSV generator, a little change on the backend is needed for this to work. [Here] the link to the backend PR](https://github.com/ProteinsWebTeam/interpro7-api/pull/183)

Also, this makes things more consistent and easier to understand: for the table, the dataKey for the short_name is counters.extra_fields.short_name, which is semantically weird.

### DIFF
--- a/src/web-workers/download/object2TSV.js
+++ b/src/web-workers/download/object2TSV.js
@@ -79,6 +79,7 @@ export const columns /*: {
   entry: [
     { name: 'Accession', selector: 'metadata.accession' },
     { name: 'Name', selector: 'metadata.name' },
+    { name: 'Short Name', selector: 'metadata.short_name' },
     { name: 'Source Database', selector: 'metadata.source_database' },
     { name: 'Type', selector: 'metadata.type' },
     { name: 'Integrated Into', selector: 'metadata.integrated' },


matthiasblum:
I'd prefer to avoid introducing API changes for this.

The client already depends on the `?extra_fields=short_name` query parameter in several parts of the codebase to include the short name in the API response.

For example, to display the InterPro entries matching `A0A000` ([link](https://www.ebi.ac.uk/interpro/protein/UniProt/A0A000/entry/InterPro/#table)), the client uses the following URL:
[`/entry/InterPro/protein/UniProt/A0A000/?extra_fields=short_name&page_size=20`](https://www.ebi.ac.uk/interpro/wwwapi/entry/InterPro/protein/UniProt/A0A000/?extra_fields=short_name&page_size=20).

We should be able to update the component(s) responsible for generating the TSV/JSON files so that `extra_fields=short_name` is also included in the entry-related URLs.

As for the way the `Table` component accesses the short name via `counters.extra_fields.short_name`, this comes from how the payload is processed:

https://github.com/ProteinsWebTeam/interpro7-client/blob/596d875034801ce0cb59f89a13594f86f3b1438d/src/components/Related/index.tsx#L72

This could be refactored to something more semantically appropriate, such as:

```ts
obj.others = omit(x, ['metadata', plural]);
```

apolignano:
Changed how the download worker makes the request to the API. Now it includes the short_name in the extra_fields.
Closing [this](https://github.com/ProteinsWebTeam/interpro7-api/pull/183).

matthiasblum (changes requested):
Unfortunately this change in `index.worker.js` works only for Entries subpages, and fail for all others. We need to add `short_name` only if the entities returned are entries.

PR #747, opened 2025-05-30
Version mismatch - JSON from IPScan 5 vs 6

apolignano:
This PR addresses [IBU-11909](https://embl.atlassian.net/browse/IBU-11909).

Now the file loader checks first if `interpro-version` is in the JSON (IPScan6), otherwise the `interproscan-version` key is considered. This is done across all the pages that perform version checks: file import, summary and status (although this last one I guess it's testable only if IPScan6 is used to run jobs)

### DIFF
--- a/src/components/IPScan/IPScanVersionCheck/index.js
+++ b/src/components/IPScan/IPScanVersionCheck/index.js
@@ -33,31 +33,25 @@ const IPScanVersionCheck = ({ data, ipScanVersion, callback } /*: Props */) => {
   if (!data || data.loading || !ipScanVersion) return <Loading inline={true} />;
 
   // If version comes from InterProScan5, split it, otherwise just take the value of "interpro-version" provided by 106
-  const jobVersion = ipScanVersion.includes('-')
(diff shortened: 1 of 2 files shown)


matthiasblum (changes requested):
A few things to address.

I tried to upload two files: 
- one generated with InterProScan 5 (with `interproscan-version` modified to trigger the mismatch warning, i.e. `5.75-106.0` instead of `5.74-105.0`)
- one generated with InterProScan 6 (also `interpro-version` modified as well, i.e. `106.0` instead of `105.0`).

## InterProScan 5

image

* The InterProScan version below "Mismatched version" can be removed as we already display the version of InterProScan just above.

## InterProScan 6

image

* The InterProScan version is not `106.0`, that's the _InterPro_ release number. The InterProScan version is `6.0.0-alpha`.
* Same comment as for InterProScan 5, the InterProScan version below  "Mismatched version" can be removed.

---

Additionally, the InterProScan and InterPro versions could be added to job details:

image

PR #748, opened 2025-05-30
EMBL Articles

apolignano:
This PR addresses [IBU-11885](https://embl.atlassian.net/browse/IBU-11885).

Now both feeds (InterPro blog and EMBL-EBI news) are retrieved, merged and ordered by date (for the InterPro blog ones string matching on the URL is used to get the date). 

Only doubts I have are: 
- should we keep displaying just 3/4 of them as it is now, or should we increase that number?
- which badge/image do we display for EMBL-EBI articles? Right now they're the same as the non-focused InterPro blog posts.

### DIFF
--- a/src/components/home/BlogEntries/__snapshots__/test.js.snap
+++ b/src/components/home/BlogEntries/__snapshots__/test.js.snap
@@ -1,30 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`<BlogEntries /> Blog entries 1`] = `
-<section>
-  <ResizeObserverComponent
-    element="div"
-    measurements={
-      [
-        "width",
-      ]
-    }
-  >
-    [Function]
-  </ResizeObserverComponent>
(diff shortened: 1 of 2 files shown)


matthiasblum (changes requested):
>     * should we keep displaying just 3/4 of them as it is now, or should we increase that number?

I think the number of blog entries to display depends on the width of the viewport:

https://github.com/ProteinsWebTeam/interpro7-client/blob/7a84d5240a5a39521170a18a2cfe3877d5c68a92/src/components/home/BlogEntries/index.tsx#L188-L190

No need to change this behaviour.

>     * which badge/image do we display for EMBL-EBI articles? Right now they're the same as the non-focused InterPro blog posts.

It's fine to use the same.

---

One thing to change, the `imageComponent` property is used to display a small label with a link to all blog posts of a given category, e.g.:

image

I think it can be removed.

matthiasblum (approved):
LGTM

PR #757, opened 2025-06-27
IPScan version check

apolignano:
This PR addresses [IBU-11962](https://embl.atlassian.net/browse/IBU-11962).

Now version mismatch warning works correctly for all the three involved sections:
- file import;
- summary;
- protein viewer;

Slight change in how the job's metadata is saved was needed.

### DIFF
--- a/src/components/IPScan/Status/SequenceList/index.tsx
+++ b/src/components/IPScan/Status/SequenceList/index.tsx
@@ -208,14 +208,18 @@ export const IPScanStatus = ({
             /> */}
         </section>
       </section>
-      <section className={css('summary-row')}>
-        <header>InterPro Version</header>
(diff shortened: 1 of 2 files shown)


matthiasblum (changes requested):
Thanks. 

Minor change to make. Also, the `status` and `expires` field in the summary could be moved above the action buttons (see screenshot)

image

apolignano:
Oops. Had tested the new logic just for imported files and not for submitted jobs. I've kept the downstream logic the same, but changed the job middleware that saves/updated the submitted jobs and related data upstream. Also, we're now temporarily hiding the IPScanVersion and IProVersion given the short delay between the "finished" status and the actual appearance of metadata and data.

matthiasblum (approved):
LGTM. Thanks!

PR #758, opened 2025-07-04
InterProScan6 download guide

apolignano:


apolignano:
Changed the page structure and now retrieving version from API. 
Make sure to add the IPScan6 param to the config file (see the example one) while testing.
Not sure if that's the most optimal position for the version, lmk.

package-lock.json
matthiasblum, line 1: Re-created with bumped version of Node / npm?

package.json
```diff
@@ -195,7 +195,7 @@
     "lint-staged": "15.2.0",
     "mini-css-extract-plugin": "2.7.6",
     "mkdirp": "3.0.1",
-    "node-sass": "9.0.0",
+    "sass": "1.89.2",
```
matthiasblum, line 198: Is this necessary for the GitHub actions to run?

matthiasblum (commented):
:cry:

image

I guess I need to install the updated dependencies. And before I do it, I'd like to be sure it's indeed necessary.

matthiasblum (changes requested):
Looks nice. Minor things to change, but in good shape.

matthiasblum (approved):
LGTM. Thanks.

PR #760, opened 2025-07-31
IPScan results minor changes

apolignano:
This PR addresses [IBU-11996](https://embl.atlassian.net/browse/IBU-11996):

- Results page:
    - imported_file prefix is now removed from the job's name. We still use it internally for other purposes;
    - No icons are now shown in the status column. Most of them are just capitalized versions of the internal status name, others are slightly modified to fit in the column nicely;
    - Tooltip has been removed from the remove button for imported files;
    - Internal ID is not shown and the link disabled while the job is being picked up by IPScan. It was also removed from the breadcrumb;

- Submission:
  I guess not the final version, but the core idea is that on the sequence search page, where you end up either by browse or when you click on a "Search sequence" button on a protein page, you do the following:
  - read the global state;
  - fetch ipscan jobs (not imported files);
  - filter the 1-sequence ones that have been submitted  in the last 60 minutes;
  - show Callout if above threshold. We have to decide threshold and better text for it.

### DIFF
--- a/src/components/IPScan/Actions/index.tsx
+++ b/src/components/IPScan/Actions/index.tsx
@@ -81,34 +81,17 @@ export const Actions = ({
   return (
     <nav className={css('buttons', { centered: forStatus })}>
       {withTitle && 'Actions: '}
-      {status !== 'imported file' ? (
-        <Tooltip
-          title={
-            <div>
-              <b>Delete job</b>: This will remove the stored data from your
-              browser. Remember that search results are only retained on our
-              servers for 7 days
-            </div>
-          }
-        >
-          {}
-          <Button
-            type={forStatus ? 'inline' : 'secondary'}
-            icon="icon-trash"
-            onClick={_handleDelete}
-            aria-label="Delete Results"
-          >
-            {!forStatus && <span>Delete Results</span>}
-          </Button>
-        </Tooltip>
-      ) : (
+      <Tooltip title={<div>Delete this job</div>}>
+        {}
         <Button
           type={forStatus ? 'inline' : 'secondary'}
           icon="icon-trash"
           onClick={_handleDelete}
           aria-label="Delete Results"
-        ></Button>
-      )}
+        >
+          {!forStatus && <span>Delete Results</span>}
+        </Button>
+      </Tooltip>
       {status === 'finished' && (
         <Tooltip
           title={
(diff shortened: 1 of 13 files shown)


src/components/IPScan/Search/index.tsx
```diff
@@ -240,6 +257,12 @@ export class IPScanSearch extends PureComponent<Props, State> {
           className={css('search-form', { dragging })}
           ref={this._formRef}
         >
+          {last60minIpScanJobs.length > 1 && (
+            <Callout type="alert">
+              {' '}
+              You've been submitting too many 1-sequence searches.
+            </Callout>
+          )}
```
matthiasblum, line 265: Don't display the callout yet. Also, when/if we display the message, we should display it closer to the submit button and be more gentle, e.g.:

> 💡 You submitted X single-sequence jobs in the past hour. You can submit up to 100 sequences in a single job to reduce queue time and server load.

matthiasblum (changes requested):
Thanks. Minor suggestions added. Also, I still see the `imported_file` prefix:

- in the URL
- in the breadcrumbs
- in the job ID

image

[results.json](https://github.com/user-attachments/files/21758305/results.json)

src/components/IPScan/Actions/index.tsx
```diff
@@ -81,15 +81,8 @@ export const Actions = ({
...
-        title={
-          <div>
-            <b>Delete job</b>: This will remove the stored data from your
-            browser. Remember that search results are only retained on our
-            servers for 7 days
-          </div>
-        }
-      >
+      <Tooltip title={<div>Delete this job</div>}>
+        {}
```
matthiasblum, line 85: Not necessary, right?

```suggestion
```

src/components/IPScan/ImportResultSearch/LoadedFileDialog/index.tsx
```diff
@@ -82,7 +82,7 @@ const LoadedFileDialog = ({
 
     saveJobInIDB(
       validFileContent.results,
-      `imported_file-${fileName}`,
+      `${fileName}`,
```
matthiasblum, line 85:
```suggestion
      fileName,
```

matthiasblum (approved):
LGTM. Two minor suggestions.

PR #769, opened 2025-09-15
TED using AlphaFold API

matthiasblum:
Switches to using the AlphaFold API instead of the TED API to retrieve TED domains, mainly so we can display the Q-score.

### DIFF
--- a/config/dev_config.yml
+++ b/config/dev_config.yml
@@ -16,7 +16,6 @@ root:
   Rfam: https://www.ebi.ac.uk/ebisearch/ws/rest/rfam
   UniProt: https://rest.uniprot.org/
   proteinsAPI: https://www.ebi.ac.uk/proteins/api/
-  ted: https://ted.cathdb.info/
 
 title: InterPro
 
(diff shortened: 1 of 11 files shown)


apolignano (commented):
Just a couple questions about boundaries:

- The link to the TED website it's still there, and the page reports different boundaries than the ones displayed. Are we fine with this inconsistency?
- Do we want all the boundaries displayed for each domain or should we display the specific one like we do for other tracks?

matthiasblum:
>     * The link to the TED website it's still there, and the page reports different boundaries than the ones displayed. Are we fine with this inconsistency?

This indeed a bit inconsistent, but TED is still the original data provider, so I think we should keep pointing to the origin of data. 

>     * Do we want all the boundaries displayed for each domain or should we display the specific ones like we do for other tracks?

If we take the example of `A0A1V6M2Y0`, the leftmost domain/location has multiple fragments/segments (97-134, 198-301, 350-375, 401-422) and it's better to show all boundaries. For matches with discontinuous domains, like `G3DSA:3.90.1150.10` in [A0A000](https://www.ebi.ac.uk/interpro/protein/UniProt/A0A000/#table), we show a thin line between individual fragments so even the tooltip only shows the boundaries of one fragment, it is clear fragments are related. We don't have that for TED, so I'd prefer to display all boundaries.

apolignano (approved):

PR #779, opened 2025-11-03
Predicted structure coloring

apolignano:
This PR addresses: [IBU-12138](https://embl.atlassian.net/browse/IBU-12138) and [IBU-12139](https://embl.atlassian.net/browse/IBU-12139)

The "Colouring theme" selector now includes representative families and domains depending on availability and it's now correctly updated when the match type (InterPro vs InterPro-N) is changed.

Changing color scheme on the protein viewer updates the color of representative data as well and that's why I find the name "theme" maybe a bit confusing. It might be worth looking for a better name (although model confidence and TED do stay the same).

The selector is now available in split screen on top of the structure.

Information about the model confidence has been moved to a tooltip next to the selector label. 
We can make descriptions for TED and representative data better.

### DIFF
--- a/src/components/ProteinViewer/Options/style.css
+++ b/src/components/ProteinViewer/Options/style.css
@@ -18,15 +18,18 @@
   .view-options-title {
     font-size: 1rem;
     font-weight: 500;
-    flex-shrink: 0;
   }
 }
 
 .view-options {
-  display: flex;
-  align-items: center;
-  justify-content: flex-end;
-  font-size: 90%;
+  display: grid;
+  grid-auto-flow: column;
+  margin-left: auto;
+}
+
+:global(.split-view) .view-options {
+  display: grid;
+  grid-auto-flow: row;
   margin-left: auto;
 }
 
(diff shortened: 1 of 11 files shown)


matthiasblum (changes requested):
Thank you.

It looks good overall, but when colouring by pLDDT score, the legend should not be hidden (and only available in a tooltip). Is it possible to display it below the structure viewer? If we need to save space, we can try to display the different items (`Very High (pLDDT > 90)`, `Confident (90 > pLDDT > 70)`, etc.) inlined instead of stacked.

> Changing color scheme on the protein viewer updates the color of representative data as well and that's why I find the name "theme" maybe a bit confusing.

Why not leaving "Colour by" like it was before?

I think I spotted a bug where colours on the domain viewer are not updated on the structure viewer:
1. open a protein, e.g. `/interpro/protein/UniProt/A0A004/alphafold/`
2. colour by TED domains
3. In the `Options` menu of the domain viewer, change the colours from `Accession` to `Domain relationship`
4. In the structure viewer, switch to colour by representative families or domains
5. The colours on the structure viewer is not the same as in the domain viewer

apolignano:
1. As discussed;
2. Changed, let me know if we want that to be the final version;
3. Turns out the bug was not from the structure color selector, but from the PV. _I guess_ that what you were trying to do was going from Domain Relationship -> Accession, or maybe the cursor got stuck on one of the tracks, but the main problem was that we've been displaying the wrong color for the representative data tracks when in ACCESSION display mode. The structure was picking up the correct color from the original representative match track (you can verify this by expanding the families section).

matthiasblum (approved):
LGTM. Made two minor commits. Feel free to merge once tests complete.

PR #787, opened 2026-01-27
UniParc page for removed UniProtKB entries

apolignano:
This PR addresses [IBU-12265](https://embl.atlassian.net/browse/IBU-12265).
A new and different component gets rendered if our API returns 404 on a UniProt accession and if UniParc has entries for it.
The new page resembles the protein one, with some changes that accommodate differences in the available data.
I think this is a good starting point. Do let me know if you want me to add/remove anything.

### DIFF
--- a/src/components/Protein/UniParcProtein/index.tsx
+++ b/src/components/Protein/UniParcProtein/index.tsx
@@ -1,35 +1,32 @@
 import React, { useRef, useEffect, useState } from 'react';
 import { createSelector } from 'reselect';
-
-import Length from 'components/Protein/Length';
+import config from 'config';
+import TitleTag from 'components/Title/TitleTag';
 import Species from 'components/Protein/Species';
 import Link from 'components/generic/Link';
 import DomainsOnProteinLoaded from 'components/Related/DomainsOnProtein/DomainsOnProteinLoaded';
 
-import { UniProtLink } from 'components/ExtLink/patternLinkWrapper';
-
-import Tooltip from 'components/SimpleCommonComponents/Tooltip';
-
 import { mergeData } from 'components/IPScan/Summary/serializers';
-import FileExporter from 'components/Matches/FileExporter';
 
 import DownloadButton from '../Sequence/DownloadButton';
-
-import { splitSequenceByChunks } from 'utils/sequence';
 
 import cssBinder from 'styles/cssBinder';
 
 import fonts from 'EBI-Icon-fonts/fonts.css';
 import ipro from 'styles/interpro-vf.css';
-import local from './style.css';
+import style from 'components/Title/style.css';
 import summary from 'styles/summary.css';
-import BaseLink from 'components/ExtLink/BaseLink';
 
-const css = cssBinder(summary, fonts, ipro, local);
+import AccessionTag from 'components/Title/AccessionTag';
+import Callout from 'components/SimpleCommonComponents/Callout';
+
+const css = cssBinder(summary, fonts, ipro, style);
 
 type Props = {
   data: {
-    metadata: ProteinMetadata & { name?: NameObject };
+    metadata: ProteinMetadata & { name?: NameObject } & {
+      taxonomies: { commonTaxon: string; commonTaxonId: string }[];
+    };
   };
   md5: string;
 };
@@ -43,9 +40,7 @@ export const UniParcProtein = ({ data, md5 }: Props) => {
 
     const fetchMatches = async () => {
       try {
-        const response = await fetch(
-          `https://www.ebi.ac.uk/interpro/matches/api/matches/${md5}`,
-        );
+        const response = await fetch(`${config.root.matches}/${md5}`);
         const data = await response.json();
         setMatches(data.matches || []);
       } catch (error) {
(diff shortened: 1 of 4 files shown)


src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+  const metadata = data.metadata;
+  const [matches, setMatches] = useState([]);
+
+  useEffect(() => {
+    if (!metadata || !metadata.accession || !md5) return;
+
+    const fetchMatches = async () => {
+      try {
+        const response = await fetch(
+          `https://www.ebi.ac.uk/interpro/matches/api/matches/${md5}`,
```
matthiasblum: Please set this URL in the config file (like `root.matchesAPI`) instead of hard-coding it here.

src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+                    </li>
+                  </>
+                ) : null}
+                {metadata.in_alphafold ? (
+                  <>
+                    <li>
+                      <BaseLink
+                        id={metadata.accession}
+                        target={'_blank'}
+                        pattern="https://alphafold.ebi.ac.uk/entry/{id}"
```
matthiasblum: Use `root.alphafold` from the config

src/pages/endpoint-page.js
```diff
@@ -140,6 +146,65 @@ class Summary extends PureComponent {
       );
     }
 
+    /* UniParc edge case */
+    if (status == 204 && payloadUniParc.results.length > 0) {
```
matthiasblum:
```suggestion
    if (status === 204 && payloadUniParc.results.length > 0) {
```

src/pages/endpoint-page.js
```diff
@@ -140,6 +146,65 @@ class Summary extends PureComponent {
...
+        },
+      };
+
+      return (
+        <>
+          <div className={f('row')}>
+            <div className={f('medium-12', 'large-12', 'columns')}>
+              <UnconnectedErrorBoundary customLocation={customLocation}>
+                <h3> UniParc Entry {standardizedData.metadata.id}</h3>
+              </UnconnectedErrorBoundary>
```
matthiasblum: Is it something that could be moved to the `UniParcProtein` component?

src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+  }, [metadata, md5]);
+
+  let organisedData = {};
+
+  if (matches && matches.length > 0) {
+    organisedData = mergeData(matches, metadata.length);
+  }
+
+  return (
+    <>
```
matthiasblum, line 62: Ideally, the title would use an `AccessionTag` and a `TitleTag` so it looks a bit like a normal protein page:

Image

`AccessionTag` can shows the UniProtKB accession, the title can stay `UniParc entry UPIxxxxxxxxx` and the `TitleTag` would be `UniParc protein`.

There would be also a callout in these lines:

> A0A4Z8KGQ7 is no longer available in UniProtKB. This page shows information from its corresponding UniParc sequence entry.

src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+              <tr>
+                <td style={{ maxWidth: '50%' }}>UniProt ID</td>
+                <td>
+                  <i
+                    className={css('shortname')}
+                    data-testid="protein-shortname"
+                  >
+                    {metadata.accession}
+                  </i>
+                </td>
```
matthiasblum: Show the UniParc ID instead

src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+              </tr>
+              <tr>
+                <td>Length</td>
+                <td data-testid="protein-length">
+                  {metadata.length} amino acids
+                </td>
+              </tr>
+              <tr>
+                <td>Taxonomy</td>
```
matthiasblum:
```suggestion
                <td>Species</td>
```

src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+                    <Tooltip title="Provided By UniProt">
+                      <span
+                        className={css('small', 'icon', 'icon-common')}
+                        data-icon="&#xf129;"
+                        aria-label="Provided By UniProt"
+                      />
+                    </Tooltip>
+                  </td>
+                </tr>
+              ) : null}
```
matthiasblum:
```suggestion
```

Never set upstream, so will always be `null`

https://github.com/ProteinsWebTeam/interpro7-client/blob/a11c6fbfe6585785bdeed4f8f3a76fe70b328387/src/pages/endpoint-page.js#L161-L174

src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+                    <li>
+                      <BaseLink
+                        id={metadata.accession}
+                        target={'_blank'}
+                        pattern="https://search.foldseek.com/search?accession={id}&source=AlphaFoldDB"
+                        className={css('ext')}
+                      >
+                        Foldseek
+                      </BaseLink>
+                    </li>
```
matthiasblum: Never set upstream

```suggestion
```

src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+        </div>
+        <div className={css('vf-stack')}>
+          <section>
+            <section>
+              <h5>External Links</h5>
+              <ul className={css('no-bullet')}>
+                <li>
+                  <UniProtLink id={metadata.accession} className={css('ext')}>
+                    UniProt
+                  </UniProtLink>
```
matthiasblum: The page pointed here will redirect to a page on UniParc, so might as well points to this page directly:

```
https://www.uniprot.org/uniparc/<uniparc-id>/entry/<uniprot-id>
```

src/components/Protein/UniParcProtein/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+              minWidth={minWidth}
+            /> 
+                // <IPScanButton
+                //     sequence={splitSequenceByChunks(
+                //         metadata.sequence,
+                //         metadata.id || '',
+                //     )}
+                //     title="Search sequence with InterProScan"
+                //     minWidth={"20"}
+                // /> */}
```
matthiasblum:
```suggestion
```

matthiasblum (changes requested):
Thanks. Looks good overall, just a few minor tweaks to make.

src/components/Protein/UniParcProtein/index.tsx
apolignano: It would be both in the title and here (?)

src/components/Protein/UniParcProtein/index.tsx
apolignano: I don't think we're sure that that's the species. For example, for Q16536, Boreoeutheria is the lowest level taxonomy id available, but it's not a species.

src/components/Protein/UniParcProtein/index.tsx
matthiasblum: Right, maybe simply remove this row then.

src/components/Protein/UniParcProtein/index.tsx
matthiasblum: Hmm, and there is actually more than one taxonomy:

- artificial sequences
- Boreoeutheria

Then instead of `Taxonomy`, use `Taxonomies`, and show all taxonomies (comma-separated).

matthiasblum (approved):
LGTM, thanks.

PR #788, opened 2026-02-11
Make observed height the default in HMM logos

apolignano:
Addresses [IBU-12327](https://embl.atlassian.net/browse/IBU-12327)


matthiasblum (approved):
LGTM. Thanks.

PR #793, opened 2026-03-02
Replace BFVD link to point to Alphafold model

apolignano:


src/components/Protein/Summary/index.tsx
```diff
@@ -284,14 +307,25 @@ export const SummaryProtein = ({ data, loading, isoform }: Props) => {
...
+                        </Link>
+                      ) : (
+                        <BaseLink
+                          id={metadata.accession}
+                          target={'_blank'}
+                          pattern={bfvdModelLinkPattern}
+                          className={css('ext')}
+                        >
+                          BFVD
+                        </BaseLink>
```
matthiasblum: What is the difference between the two links?

Also, now that we redirect the the AlphaFold website, the text of the link should be "AlphaFold DB" instead of BFVD.

matthiasblum (changes requested):
Thanks.

The text of the external link should reflect the target (see suggestion below).

I think something is wrong with the pLDDT score displayed in the domain viewer. Example for `A0A8S5SFL1`:

Image

The pLDDT scores are very high, so the pLDDT track should be mostly dark blue. The same happens in the live website: https://www.ebi.ac.uk/interpro/protein/UniProt/A0A8S5SFL1/

Finally, in the "BFVD" subpage, the link "View on BFVD" should also be updated.

Image

apolignano:
A few things:

- I've refactored the code and changed some pieces of logic. The additional API call to get the name of the model was not necessary. The link just needs the accession as it did for BFVD and as it does for all the "normal" AFDB entries.
- The difference between the two links was that one was the fallback just in case the API call failed/the entry was not available, but the pattern indeed needed some adjustment. Now that distinction is not needed anyway.
- The problem with the plddt track is that we never displayed the BFVD confidence score on the Overview page, but only on the BFVD tab. Now that BFVD models and confidence scores are available on AFDB they are being picked up by an API call. Problem is, the models are different (a "v2" instead of "v6", but I guess that's specific to BFVD) and the confidence categories as well (the letters for coloring the Nightingale track). I'll open a different PR to address this problem.

matthiasblum (approved):
LGTM. Thank you.

PR #796, opened 2026-04-09
Multimers, no BFVD

apolignano:


src/components/Structure3DModel/3DModel/index.tsx
```diff
@@ -129,94 +166,60 @@ const Structure3DModel = ({
...
-              </>
-            )}
-            .
+            The protein structure below has been predicted by{' '}
+            <Link href={'//deepmind.google/'}>Google DeepMind</Link> using
+            AlphaFold (
+            <Link href={'//www.nature.com/articles/s41586-021-03819-2'}>
+              Jumper, J et al. 2021
+            </Link>
+            )
```
matthiasblum: I would suggest removing this block. We initially added on day-1 of the release of AlphaFold (and the publication of the paper), but with BFVD predictions being added to AlphaFold DB this could be misleading because BFVD structures were not predicted with AlphaFold 2.

src/components/Structure3DModel/AlphaFoldStructuresTable/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+            cursor: 'pointer',
+            border: 'none',
+            background: 'none',
+            color: 'inherit',
+            textDecoration: 'underline',
+            padding: 0,
+            fontSize: '0.85em',
+          }}
+        >
+          [{isVisible ? 'Hide' : 'Show'}]
```
matthiasblum: Maybe use `> Expand` and `v Collapse` instead of `Show` and `Hide`, using the `chevron circle` icons (https://www.ebi.ac.uk/style-lab/general/fonts/v1.3/)

src/components/Structure3DModel/AlphaFoldStructuresTable/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+                onClick={() => onSelect(doc.entryId)}
+                style={{
+                  cursor: 'pointer',
+                  fontWeight: doc.entryId === selectedId ? 'bold' : 'normal',
+                }}
+              >
+                <td>{doc.entryId}</td>
+                <td>{doc.uniprotStart}</td>
+                <td>{doc.uniprotEnd}</td>
+                <td>{doc.provider || 'AlphaFold'}</td>
```
matthiasblum: I see a `providerId` property, but no `provider`. Additionally, I'd rename the column header to _Dataset_ and use the same name as used by AlphaFold on their website (e.g. `GDM` -> `Google DeepMind`)

matthiasblum (commented):
Thanks.I haven't reviewed everything yet, but I noticed that for proteins with BFVD prediction, the AlphaFold tab is missing.

For instance, for `P21405`:

Image

Opening `/interpro/protein/UniProt/P21405/alphafold/` shows everything as expected, but it would be nice for `/interpro/protein/UniProt/P21405/bfvd` to redirect to the `alphafold` subpage.

apolignano:
Addressed all the comments.
The problem with the AlphaFold tab not showing is probably caused by the API version/db that you're using to test. That tab is affected by the in_bfvd/in_alphafold recent change (see below).

image

matthiasblum:
> Addressed all the comments. The problem with the AlphaFold tab not showing is probably caused by the API version/db that you're using to test.

Which API branch should I use to test, then?

apolignano:
Oops, sorry, I thought I had already opened the PR for that but I hadn't published the branch yet. It's the `remove_bfvd` one.

src/components/EntryMenu/EntryMenuLink/index.tsx
```diff
@@ -88,13 +88,7 @@ export const EntryMenuLink = ({
       name.toLowerCase() === 'alphafold'
     ) {
       value = (payload.metadata as EntryMetadata).in_alphafold ? 1 : 0;
```
matthiasblum: Counter next to AlphaFold is set to `1` even if there are multiple predictions linked to the protein. Is possible to update the counter once we have fetched the list of available predictions from the AlphaFold API? If it's not possible, I'd suggest hiding the counter (for protein pages only, for entry pages, it's coming from the API).

src/components/Structure3DModel/3DModel/index.tsx
```diff
@@ -390,55 +389,90 @@ const Structure3DModel = ({
...
+          modelsUrl = format({
             protocol,
             hostname,
             port,
             pathname: isUrlToApi
-              ? `${root}api/prediction/${accession}`
+              ? `${root}api/search`
               : `${root}entry/${accession}`,
+            search: isUrlToApi
+              ? `?q=(uniprotAccession:${accession})&type=main`
```
matthiasblum:
```suggestion
              ? `?q=(uniprotAccession:${accession})&type=main&start=0&rows=100`
```

The AlphaFold API returns the first 10 predictions by default. If there are more than 10 predictions, the text above the table shows an incorrect count (_Available structures (10)_). As of today, the UniProt accession with the highest number of predicted structures is `P0C6X2` (51 entries), so adding `rows=100` should solve this issue.

src/components/Structure3DModel/3DModel/SequenceCheck/index.tsx
matthiasblum, line 1: This has odd consequences long sequences. 

`P0C6X2` has predicted structures on chunks of its sequence only. So all structures, including the one selected by default, have the _AlphaFold Prediction Mismatch_ warning... with a huge diff.

Maybe we could remove the diff and simply the warning message to something like this:

> This structure prediction was generated from a sequence that differs from the current UniProt sequence. The displayed structure may therefore be inaccurate or incomplete.

src/components/Structure3DModel/AlphaFoldStructuresTable/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const AlphaFoldStructuresTable = ({ docs, onSelect, selectedId }: Props) => {
+  const [isVisible, setIsVisible] = useState(true);
+
+  if (docs.length === 0) return null;
+
+  const sortedDocs = [...docs].sort((a, b) => {
+    const providerA = (a.provider || 'AlphaFold').toLowerCase();
+    const providerB = (b.provider || 'AlphaFold').toLowerCase();
```
matthiasblum:
```suggestion
    const providerA = (a.providerId || 'AlphaFold').toLowerCase();
    const providerB = (b.providerId || 'AlphaFold').toLowerCase();
```

src/components/Structure3DModel/AlphaFoldStructuresTable/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+                  onClick={() => onSelect(doc.entryId)}
+                  className={css({
+                    'structure-row': true,
+                    selected: doc.entryId === selectedId,
+                  })}
+                >
+                  <td>{doc.entryId}</td>
+                  <td>{doc.uniprotStart}</td>
+                  <td>{doc.uniprotEnd}</td>
+                  <td>{doc.provider || 'AlphaFold'}</td>
```
matthiasblum:
```suggestion
                  <td>{doc.providerId || 'AlphaFold'}</td>
```

src/types/external-payloads.d.ts
```diff
@@ -157,22 +157,55 @@ type RfamPayload = {
...
+
+type AlphafoldSearchDoc = {
+  entryId: string;
+  uniprotAccession: string;
+  uniprotStart: number;
+  uniprotEnd: number;
+  oligomericState: string;
+  assemblyType?: string;
+  latestVersion: number;
+  provider?: string;
```
matthiasblum:
```suggestion
  providerId?: string;
```

src/types/external-payloads.d.ts
```diff
@@ -157,22 +157,55 @@ type RfamPayload = {
...
   uniprotId: string;
   uniprotDescription: string;
   taxId: number;
   organismScientificName: string;
+  isComplex: boolean;
+  uniprotStart: number;
+  uniprotEnd: number;
+  oligomericState?: string;
+  latestVersion?: number;
+  provider?: string;
```
matthiasblum, line 188:
```suggestion
  providerId?: string;
```

matthiasblum (changes requested):
Thanks. I found a few issues that are either reported below or directly in relevant files:

### AlphaFold structure not updated on entry pages

1. Head to `/entry/InterPro/IPR000001/alphafold`
2. The first structure is `R7T3V2`. It has only one prediction. Everything works as expected.
3. Select another protein, e.g. `A0A8W8NPB7`: the table is updated and list the predictions available for `A0A8W8NPB7`, but the structure displayed is not updated and the _AlphaFold Prediction Mismatch_ warning appears. The user needs to select a prediction from the table to update the structure viewer. Instead, we should select a prediction and display it.

### Outdated text

The following text is displayed on all AlphaFold pages, but it's not necessarily try:

> The protein structure below has been predicted by [Google DeepMind](http://deepmind.google/) using AlphaFold ([Jumper, J et al. 2021](http://www.nature.com/articles/s41586-021-03819-2))

I would simply remove it. We already provide other links to AlphaFold.

src/components/Structure3DModel/3DModel/SequenceCheck/index.tsx
apolignano: Oh good catch. Now also sorting by sequence start.

matthiasblum (changes requested):
For the AlphaFold counter, it's set to `1` then updated when navigating to the `AlphaFold` subpage. For instance with `P0C6X2`:

Overview:

Image

AlphaFold:

Image

Is it possible to set the correct/final count on any subpage for the protein? If it's not, then I'd rather have no count at all than an inexact one. It could look like this:

Image

---

Heterodimers are available in the AlphaFold API, but some of the properties we rely on can be `null` (e.g. `uniprotStart`, `uniprotEnd`). In the case of `C1HCX1`, the heterodimer prediction is selected by default:

Image

When a GDM/monodimer and a NVDA/heterodimer are available, can you make sure the former is selected by default?

matthiasblum (approved):
Thank you. LGTM.

PR #799, opened 2026-05-18
Add CAPTCHA for ticket submission

apolignano:
This PR addresses https://embl.atlassian.net/browse/IBU-12438.

An HCaptcha React component now shows up on the feedback page. If validated, the submit button appears.

To address:
- I'm the one who created the HCaptcha account. It might be worth discussing how to have that account registered under a team/group email.

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -1,4 +1,6 @@
 googleAnalytics: UA-XXXXXXXX-X
+hcaptchaSiteKey: hcaptcha-sitekey
+
 root:
   website: http://localhost:8080/interpro/
   API: https://www.ebi.ac.uk/interpro/api/
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
       "dependencies": {
         "@floating-ui/react": "0.26.5",
         "@floating-ui/react-dom": "2.0.5",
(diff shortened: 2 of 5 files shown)


matthiasblum (changes requested):
Looks good overall, with one thing that needs to be changed.

At first, I ran the API and the client on my laptop, without being connected to the VPN (MySQL and Elastic connections going through SSH tunnels). Not being on the EBI network, the email failed to be sent to `smtp.ebi.ac.uk`: the following block raised an exception:

https://github.com/ProteinsWebTeam/interpro7-api/blob/2a78f22bea7353eb2db8e5d6ab6d7cd9db50944b/webfront/views/mail.py#L39-L46

Since it's not caught, the API returns a 500 and an "Invalid request" callout is displayed:

https://github.com/ProteinsWebTeam/interpro7-client/blob/e5a65ee2856dbaafeb66a416c97e009544128258/src/subPages/Feedback/index.tsx#L90-L91

Would it be possible to show an more informative error message, e.g.

> Sorry, we couldn't send your feedback. You can submit it through our [contact form](https://www.ebi.ac.uk/about/contact/support/interpro) instead. Please mention that you encountered a sending error.

matthiasblum (approved):
LGTM. Thanks!

PR #802, opened 2026-06-04
Update InterProScan download page

matthiasblum:
- Move the InterProScan 5 section _after_ InterProScan 6
- Add a warning about the end-of-support of InterProScan 5
- Fetch the current InterPro version for the InterProScan 6 command

### DIFF
--- a/src/components/IPScan/DownloadTable/index.tsx
+++ b/src/components/IPScan/DownloadTable/index.tsx
@@ -72,7 +72,7 @@ export const DownloadTable = ({
       </p>
       <h4>InterProScan 5</h4>
       <Callout type="warning">
-        InterProScan {interProScan5Version} is the finale release of
+        InterProScan {interProScan5Version} is the final release of
         InterProScan 5. To annotate sequences with future InterPro releases,
         please upgrade to InterProScan 6.
       </Callout>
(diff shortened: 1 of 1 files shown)


src/components/IPScan/DownloadTable/index.tsx
```diff
@@ -20,20 +21,61 @@ type ReleasePayload = Array<{
...
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          GitHub
+        </a>
+        .
+      </p>
+      <h4>InterProScan 5</h4>
+      <Callout type="warning">
+        InterProScan {interProScan5Version} is the finale release of
```
apolignano:
```suggestion
        InterProScan {interProScan5Version} is the final release of
```

src/components/IPScan/DownloadTable/index.tsx
```diff
@@ -81,38 +123,6 @@ export const DownloadTable = ({ dataIPScan, dataIPScan6 }: LoadedProps) => {
         </Link>
```
apolignano, line 123: Maybe we could move this bit out of this little small `<p>` and have it as a normal size `<p>` like the IPScan6 section above? We can leave the rest of the FTP bit as is. Seems more consistent.

apolignano (approved):
LGTM. Just two minor comments.

src/components/IPScan/DownloadTable/index.tsx
matthiasblum: Hmm, the link to the repo is not that important (since we don't expect users to build from the code). Not changing this.

PR #803, opened 2026-06-10
IPScan6 paper and social links

apolignano:


src/components/Help/index.tsx
```diff
@@ -255,15 +255,25 @@ const Help = () => {
                 withReferrer
               >
                 <span
-                  className={css('icon', 'icon-common')}
+                  className={css('icon', 'icon-common', 'icon-medkit')}
                   data-icon="&#x6e;"
```
matthiasblum:
```suggestion
```

No need of `data-icon` if using the `icon-medkit` class.

src/components/Help/index.tsx
```diff
@@ -293,7 +303,7 @@ const Help = () => {
                   className={css('icon', 'icon-common')}
                   data-icon="&#x6e;"
```
matthiasblum: Please also add the medkit icon here.

src/components/Twitter/index.tsx
```diff
@@ -26,7 +26,7 @@ const Twitter = ({ handler = 'InterProDB' }: Props) => {
         target="_blank"
       >
         {' '}
-        Posts by @{handler}
+        X (Twitter)
```
matthiasblum:
```suggestion
        Twitter
```

I'd stand by _Twitter_ only, as the icon is already an _X_.  If you strongly disagree, then I'd suggest:

```suggestion
        @{handler}
```

(A handle is always better than a single letter).

matthiasblum (approved):
Thanks. Minor change to make so the icons of the "Heldesk" links are consistent.

src/components/Twitter/index.tsx
apolignano: Yeah, not Twitter anymore 😆 
Let's use the handle.

PR #804, opened 2026-06-10
Heterodimers coloring fix

apolignano:


src/components/Structure3DModel/selectors.ts
```diff
@@ -35,12 +51,16 @@ export const getConfidenceURLFromPayload = (namespace: string) =>
...
-      const cifURL =
-        selectedCifUrl ||
-        dataPrediction?.payload?.find(
-          (item) =>
-            item.uniprotAccession === ((search.isoform as string) || accession),
-        )?.cifUrl;
+      const matchedCifUrl = dataPrediction?.payload?.find(
+        (item) =>
+          item.uniprotAccession === ((search.isoform as string) || accession),
+      )?.cifUrl;
```
matthiasblum: I think this is what causes the bug reported in https://helpdesk.ebi.ac.uk/Ticket/Display.html?id=874775. We expect the payload to be an array of objects, but the payload of `https://alphafold.ebi.ac.uk/api/prediction/P51587` _is_ an object, so `find()` doesn't exist.

src/components/Structure/Viewer/index.tsx
matthiasblum, line 141: I missed this during a previous review. What is that?

Following this change, the prediction never loads when looking at the list of proteins with AlphaFold predictons (e.g. `/interpro/entry/InterPro/IPR000001/alphafold`)

matthiasblum (changes requested):
Requires changes or clarifications.

src/components/Structure3DModel/selectors.ts
apolignano: Unfortunately it seems it can be both. Empty object or list with objects. Fixing typing and selector.

src/components/Structure/Viewer/index.tsx
apolignano: I don't think that's what's causing it. That function and _pdbe_ url have been in the viewer for a while, but recently moved around when coloring/multimer changes were introduced. The fix here is to make the Entry page use the creators and filter chain already introduced in the protein page.

matthiasblum (approved):
LGTM, thanks!

PR #806, opened 2026-06-17
hCaptcha token, host and browser errors

apolignano:
This PR fixes some hCaptcha problems:
- On VMs/K8s, we'll specify the host;
- If ad blockers/extensions prevent hCaptcha loading, a warning message appears;
- The client gets a response from the usual /mail endpoint, but handles errors coming from the secret verification now happening on the backed (see https://github.com/ProteinsWebTeam/interpro7-api/pull/212).


src/subPages/Feedback/index.tsx
matthiasblum, line 46: Missed this in a previous PR, but shouldn't the hooks be defined before an early return?

matthiasblum (commented):
There seems to be an issue with the hCaptcha widget, but I am not sure whether it's code-related or deployment-related.

Basically, when I click on "I am human", this warning is displayed above the form:

Image

When clicking a second time, the challenge correctly appears:

Image

Then I can complete the challenge and the feedback is successfully sent. But we shouldn't have to click twice on the checkbox to display the challenge.

Another idea, related to the position of the challenge (far right of the page): right now completing the challenge shows the submit button only. How difficult would it be to make it show the full form (email, feedback, submit button)?

matthiasblum (approved):
LGTM, thanks!

One minor thing that could be done, but no big deal if not done in this PR, is to hide the hCaptcha widget once the feedback has been successfully submitted:

1. At this point, the user already passed the hCaptcha challenge, they should not have to prove they are human again
2. And I think it's better to hide the form as it gives a clean signal that the task (sending the feedback) is done

PR #808, opened 2026-07-16
Add InterProN matches and extra features to .tsv download on Protein pages

apolignano:
Addresses [IBU-12544](https://embl.atlassian.net/browse/IBU-12544).

- Upon requesting the generation of the .tsv, the already cached extra data (intepro_n and extra_features are requested on load) gets fetched;
- The fetched data gets processed by the already existing .tsv processing functions, using the same columns we had before (we use the db column to identity InterPro-N matches); 
- Sligthly improved UI style for the three (search, download fasta, download file) buttons.

### DIFF
--- a/src/components/Protein/Summary/index.tsx
+++ b/src/components/Protein/Summary/index.tsx
@@ -11,7 +11,10 @@ import Link from 'components/generic/Link';
 import ProteinEntryHierarchy from 'components/Protein/ProteinEntryHierarchy';
 
 import { UniProtLink } from 'components/ExtLink/patternLinkWrapper';
-import DomainsOnProtein from 'components/Related/DomainsOnProtein';
+import DomainsOnProtein, {
+  getExtraURL,
+  getInterProNMatches,
+} from 'components/Related/DomainsOnProtein';
 
 import loadable from 'higherOrder/loadable';
 import {
@@ -40,7 +43,9 @@ import local from './style.css';
 import summary from 'styles/summary.css';
 import BaseLink from 'components/ExtLink/BaseLink';
 
-const css = cssBinder(summary, fonts, ipro, local);
+import buttonCSS from 'components/SimpleCommonComponents/Button/style.css';
+
+const css = cssBinder(summary, fonts, ipro, local, buttonCSS);
 
 const SchemaOrgData = loadable({
   loader: () => import(/* webpackChunkName: "schemaOrg" */ 'schema_org'),
@@ -77,9 +82,17 @@ type Props = {
   loading: boolean;
 };
 
-interface LoadedProps extends Props {}
+interface LoadedProps
+  extends Props,
+    LoadDataProps<ExtraFeaturesPayload, 'Features'>,
+    LoadDataProps<InterProNMatches, 'InterProNMatches'> {}
 
-export const SummaryProtein = ({ data, loading }: LoadedProps) => {
+export const SummaryProtein = ({
+  data,
+  loading,
+  dataFeatures,
+  dataInterProNMatches,
+}: LoadedProps) => {
   const comparisonContainerRef = useRef<HTMLElement | null>(null);
   const [matchesLoaded, setMatchesLoaded] = useState(false);
   const [families, setFamilies] = useState<Array<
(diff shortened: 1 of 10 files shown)


src/web-workers/download/index.worker.js
```diff
@@ -221,6 +228,7 @@ const downloadContent =
           version = response.headers.get('InterPro-Version');
           mutatePayloadTo3rdPartyAPI(payload, endpoint, firstPage);
           totalCount = payload.count;
+          if (!firstProtein) firstProtein = payload.results?.[0]?.proteins?.[0];
```
matthiasblum: What happens if the sequence has no matches?

src/web-workers/download/index.worker.js
```diff
@@ -243,6 +251,27 @@ const downloadContent =
...
+            extraData.interpro_n,
+            firstProtein?.accession,
+            firstProtein?.protein_length,
+          ),
+          ...extraFeaturesToRows(
+            extraData.extra_features,
+            firstProtein?.accession,
+            firstProtein?.protein_length,
+          ),
+        ];
```
matthiasblum: The `Protein Accession` and `Protein Length` columns could be empty if the protein is only matched by extra features or InterPro-N.

src/components/Protein/Summary/index.tsx
```diff
@@ -294,6 +311,7 @@ export const SummaryProtein = ({ data, loading }: LoadedProps) => {
                 label="Download matches (TSV)"
               />
             </label>
+            {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
```
matthiasblum: Should probably be _above_ the `<label>` line.

src/components/Protein/Summary/index.tsx
```diff
@@ -316,4 +334,12 @@ export const SummaryProtein = ({ data, loading }: LoadedProps) => {
...
 };
 
-export default SummaryProtein;
+export default loadData<ExtraFeaturesPayload, 'Features'>({
+  getUrl: getExtraURL('extra_features'),
+  propNamespace: 'Features',
+} as LoadDataParameters)(
+  loadData<InterProNMatches, 'InterProNMatches'>({
+    getUrl: getInterProNMatches,
+    propNamespace: 'InterProNMatches',
```
matthiasblum: Isn't that already done in `DomainsOnProtein`?

matthiasblum (changes requested):
One edge case that could lead to column with missing values (sequence without "traditional" matches but extra matches and/or InterPro-N matches). I don't have an example accession at hand, so you could emulate an empty response for trad matches.

Also, thanks for improving the UI of the buttons. Instead of "Generate TSV", would it be possible for the label to be "Download matches (TSV)"?

src/components/Protein/Summary/index.tsx
apolignano: Yes and although I thought that caching would save us from doing duplicate requests on a cold load, it's not the case. Apparently the first load does duplicate requests anyway. Cache works from the second load on. Fixed this with a new strategy.

src/components/Related/DomainsOnProtein/index.tsx
```diff
@@ -274,7 +289,7 @@ const getRelatedEntriesURL = createSelector(
   },
 );
 
-const getExtraURL = (query: string) =>
+export const getExtraURL = (query: string) =>
```
matthiasblum:
```suggestion
const getExtraURL = (query: string) =>
```

src/components/Related/DomainsOnProtein/index.tsx
```diff
@@ -325,7 +340,7 @@ const getPTMPayload = createSelector(
   },
 );
 
-const getInterProNMatches = createSelector(
+export const getInterProNMatches = createSelector(
```
matthiasblum:
```suggestion
const getInterProNMatches = createSelector(
```

src/web-workers/download/index.worker.js
```diff
@@ -217,10 +239,18 @@ const downloadContent =
...
+          // No content: this page has no results (e.g. a protein with no
+          // standard matches), so there's nothing to parse as JSON.
+          if (response.status === NO_CONTENT) {
+            next = null;
+            errorCount = 0;
+            totalCount = extraRows.length;
+            continue;
+          }
           const payload = await response.json();
           version = response.headers.get('InterPro-Version');
```
matthiasblum: We should probably get the version even if the response is 204. I imagine that having `version` left to `null` may lead to undesired downstream behaviours.

matthiasblum (changes requested):

matthiasblum (approved):

PR #812, opened 2026-08-03
Restore representative structure images

apolignano:
Addresses [IBU-12570](https://embl.atlassian.net/browse/IBU-12570).

- We now show the three structure pictures on the entry page in the "Representative structure" section;
- There were other pages (structure browse, structure matches) on InterPro that were using PDBSum. They're now fetching the structure from PDBe (single image) as well.
- Removed all the links and references to PDBSum;


matthiasblum (approved):
LGTM. One minor improvement (but not blocking merging this PR) would be to create a component wrapping `LazyImage` and taking the params: the PDB ID, and `front`, `side`, or `top`. This component would render the requested image for the PDB ID of interest. This would allow to have one single place with the PDBe static URL hardcoded.

PR #813, opened 2026-08-05
Docbot

apolignano:
Added Docbot.

Branched docbot out of master, then merged to dev to test it out on [int](https://wwwint.ebi.ac.uk/interpro/), then merged dev back into docbot, so that we're up to date with the rest as well. (There are updates for IPScan6, so tell me if you're ok with that).

Please let me know once you've tested it so that I'll do some final changes and edit the config accordingly.

Thanks!


matthiasblum (changes requested):
Works fine on https://wwwint.ebi.ac.uk/interpro. 
Only change needed is to make sure that the `border-radius: 50%;` style on `.vf-chatbot-fab` is not superseded by global styles on `[type="button"]`.

matthiasblum (approved):

